PR #24, opened 2014-01-25
Fixes for using option values inside query expressions.

janno-p:
Found some more bugs:

When query expression created projection in which only option types were selected, the underlying SQL would instead query all columns from database. Because of that behavior, `distinct` queries were not working for option types.

Another problem was with retrieving `NULL` values. Because Npgsql driver returns `DBNull` instead of `null`, some nulls were getting through and causing `InvalidCastException` in `SqlEntity::GetColumnOption<'T>(key)` method.


Thorium:
Workaround given. Closed as answered without follow-up.

PR #81, opened 2014-07-13
Update SQLite provider to work with Mono.

rachelblasucci:


bonsmik:
@rachelreese any simple examples how to get things working on mono? Have been trying to get it working on osx without any luck.

rachelreese:
@bonsmik Apologies for the slow response. :) You could check out https://github.com/rachelreese/Tasky but I've not updated the version of the SQLProvider that I use on the project in ages, and the syntax might be out of date. It's on my to-do list, but it might be a couple days before I have the chance to properly check and update that. Let me know, though!

Issue #19, opened 2014-01-22
The world needs more type providers 


forki:
Instead of making the database provider a static parameter of a god type provider consider to create specialized type providers for every database vendor. 

This would make it easier to match optional parameters for the concrete provider (at least for the Dynamics NAV provider)

Thorium:
Done.

Thorium:
Hmm. We could do quite easily support for more ADO.NET / SQL -style providers:

- OLAP, supports ADO.NET and MDX is enough near SQL. Would be new domain to F# type providers? Also, interesting: e.g. https://bennyaustin.wordpress.com/2011/03/01/ssas-dmv-queries-cube-metadata/ 
- No-one uses DB2, right?

Odbc covers many cases but Odbc configs are a little bit clumsy. Then these more dynamic data sources, but maybe they could fetch design time schema data better with some independent type provider, and ODATA type provider already exists:
- Elasticsearch: Could be queried with SQL? https://github.com/NLPchina/elasticsearch-sql 
- MongoDB: No ADO but this: https://github.com/ElemarJR/LinqToMongo/tree/master/src although there is already easy non-TP way (maybe more natural?) to use mongo from F#
- Redis...?

PR #85, opened 2014-07-27
The God Provider Is Not Enough

simonhdickson:
For #19

Separates out the the god Type Provider into multiple. All the parts I've tested seem to work (I don't have most of the DBs set up locally), very little actual code is changed though. If everyone can check that everything still works and give any feedback that would be great.


colinbull:
What should we do with this? I guess we can scrap it, as it has probably diverged to much?

pezipink:
yeah it's too far out of sync now.   revist this later.

Thorium:
Now that Microsoft has marked System.Data.SqlClient as obsolete and replacing it with Microsoft.Data.SqlClient, which has references to half of the internet (like nodemodules), I think it's time to revisit this, splitting the current provider to multiple NuGet packages with separate dependencies. We can even keep the current god-provider as "common" for backward compatibility and add these database-dependant ones to reference the common module.

PR #113, opened 2015-01-30
Convert to paket and latest ProjectScaffold

forki:
This converts the SQLProvider project to the lastest and greatest from ProjectScaffold.

Release process see http://fsprojects.github.io/ProjectScaffold/release-process.html


colinbull:
Merged into #127. Thanks

PR #116, opened 2015-02-06
Fix array for IN and NOT IN expressions in PSQL. Fixed generic calls.

netogallo:
I tried running the test suite: PostgreSQLTests and failed for two errors.

1) Whenever the 'IN' or 'NOT IN' operators where used, the PSQL provider created a parameter having the whole array as parameter instead of separating the array into multiple parameters. I fixed this by checking what the operation is when pattern matching since the IN and NOT IN operators always take an array as argument.

2) The SetColumns method was not being treated as a generic method by .Net reflection (don't know about Mono) so I added an explicit generic parameter to avoid confusion.


pezipink:
Great! Thanks.  I remember this issue a while back in the other providers.

Issue #93, opened 2014-09-19
UseOptionTypes=true causes generate writable properties of type 'a option option


vasily-kirichenko:
```
The type provider 'FSharp.Data.Sql.SqlTypeProvider' reported an error in the context of provided type 'FSharp.Data.Sql.SqlDataProvider,ConnectionString="Data Source=test_server; Initial Catalog=TEST; Integrated Security=True",UseOptionTypes="True"+dataContext+[dbo].[test]Entity', member 'set_NAME'. The error: Type mismatch when building 'args': invalid parameter for a method or indexer property. Expected 'Microsoft.FSharp.Core.FSharpOption`1[Microsoft.FSharp.Core.FSharpOption`1[System.String]]', but received type 'Microsoft.FSharp.Core.FSharpOption`1[System.String]'.
Parameter name: receivedType	L:\stash\...\Source\Tests\DB.fs	34	5	Tests
```
I expect the type of `set_NAME` to be `string option` instead of `string option option`.

Issue #119, opened 2015-02-19
Option types


jamessdixon:
I swapped out the nuget version with the latest from master and my option type assignments are now failing.

If I have a nullable varchar field in the database and try and assign a string option, I am seeing:

The error: Type mismatch when building 'args': invalid parameter for a method or indexer property. Expected 'Microsoft.FSharp.Core.FSharpOption`1[Microsoft.FSharp.Core.FSharpOption`1[System.String]]', but received type 'Microsoft.FSharp.Core.FSharpOption`1[System.String]'.

PR #120, opened 2015-02-21
Fix option type properties (issue #119).

janno-p:
Generic type argument for `GetColumnOption` and `SetColumnOption` methods should be underlying type not the `option<_>` type itself.


janno-p:
Actually it should fix #93 too, because they're essentially same with #119.

pezipink:
I was just looking at this myself!  I have no idea how this happened as the option types were ceriianly working properly at some point, the provider has been around a long time now.

PR #124, opened 2015-03-06
Bug: returned commands were already disposed

orient-man:
SubmitUpdates for SQLite didn't work at all: created command were disposed before execution


pezipink:
thanks for this, not sure how that slipped through!!

Issue #94, opened 2014-09-19
Everything is in upper case (MSSQL)


vasily-kirichenko:
I built it from source and it generates everything in upper case now (the version that lays in NuGet generates normal names).

Is it intentional?

Thorium:
+1
Can't wait to get this from NuGet. ;-)

colinbull:
Yeah, sorry!! we keep saying we will get a release out soon. But I have been super busy for what seems like forever, on C# projects.

pezipink:
Sorry, yes, I am terrible as well.  We have some proper docs being written and I am intending to spend some time with this project soon and get it sorted out, it's been a year since the last package (!!)

Thorium:
After #143 you can now do this:

    open FSharp.Data.Sql.Common
    type sql = 
      SqlDataProvider<(* ... *), CaseSensitivityChange=CaseSensitivityChange.ORIGINAL>

Issue #87, opened 2014-08-01
Nicer naming for DB objects.


colinbull:
I have had a few requests from people I know that are using it (as well as FSharpComposableQuery support (issue #48 and [here](https://github.com/fsprojects/FSharp.Linq.ComposableQuery/issues/7#issuecomment-50934573)) , about cleaning up the naming. So things like backticks and `[]` are minimized, basically so it looks like normal properties. Additionally, this may also include removing `_` and case names properly.

On an implementation note, this doesn't look to bad as we can retain the DB name in a type that represents the DB object name.

Issue #48, opened 2014-03-15
Will this provider work with FSharp.Linq.Experimental.ComposableQuery?


ovatsus:
I assume no because it's erased instead of generated, but maybe there can be a way to make it work?
http://fsprojects.github.io/FSharp.Linq.Experimental.ComposableQuery/

Thorium:
Closed by recommendation of Copilot.

PR #127, opened 2015-03-12
[WIP] Consolidating pull requests

colinbull:
This is a first cut of the nicer naming feat, still need to fix up the main docs, but that needs doing in general :).


forki:
that's looking really cool!

colinbull:
@forki :+1:

colinbull:
Should note this basically attacks issues #48, #87 and #94

pezipink:
awesome!

Thorium:
This PR has over 50 commits and there are some breaking changes. How about merging it to a different branch and releasing it to nuget as a pre-version (alpha/beta)? Oh..but the current is also alpha...0.0.9?

This branch is auto-running some unit tests that other branches are not. Those unit tests have file paths in [<Literal>] ConnectionString "data source" and ResolutionPath, and they use backslashes so that is though to the Travis CI (Mono). 

Some foreign keys have FK_Thing_0_0 style names. What are these 0 and 0? In Mono SQLite they are just FK_Thing_0

I don't know if it's just me but my unit tests in Windows are failing because of MissingMethodException and this:

     <@@ match ConfigHelpers.tryGetConnectionString true runtimePath conStringName connnectionString with
         | Some(cs) -> cs
         | None -> failwithf "No connection string specified or could not find a connection string with name %s" conStringName @@>

If I replace it with this then everything is ok:

     <@@ connnectionString @@> 

The reason seems to be that nunit.framework seems to do some failing lookup before the actual runtime execution. So it should be pretty hard to solve this with runtime code.

colinbull:
@pezipink  @Thorium O.K. after merging master back into this P.R. which was kinda epic. I'm beginning to get this P.R. back on its feet (apart from the build, which I'll sort out last). There quiet a few conflicting changes so I'm hoping I haven't messed up to much in resolving them. But it seems to stand up and does  build and run on both mono (4.0.2) and windows. 

Also not that I have changed the naming on the relationships. This is mainly due to the providers providing different names depending on the platform. So a query using relationship properties now looks like this 

     let query = 
        query {
            for cust in dc.Main.Customers do
            for order in cust.``main.Orders by CustomerID`` do
            for orderDetail in order.``main.OrderDetails by OrderID`` do
            select (cust.CustomerId, orderDetail.ProductId, orderDetail.Quantity)
        } |> Seq.toArray

This is just a first stab, any suggestions are welcome. 

So do we want to now go ahead and pull this PR in to the repository but on a `dev` branch?

pezipink:
Great, I will try this out my side.  I think the relationships are better the way you are doing them now anyways :)

I reckon we can just merge this into main when happy, this is currently pre-release so breaking changes are ok heading to the V1, I think there are already some from ages ago anyway.

Thorium:
1) I can't get MySQL-provider to find table column names. Something with the 
member __.GetColumns(con,table) or columnLookup-cache goes wrong. However MySqlTests.fsx works fine. The only real difference is that I have schema and table in lower case, and I have column names in CamelCase.

2) Test projects:

MySqlTests.fsx, line 54: emp.employees_ibfk_3 don't exist.
 emp.``hr.employees by MANAGER_ID``

SqlServerTests.fsx, line 77: emp.EMP_MANAGER_FK don't exist.
emp.``dbo.Employees by EMPLOYEE_ID_0``

SqlServerTests.fsx, lines 159,174,177,186: ResultSet don't exist.

QueryTests.fs and CrudTests.fs, have literal that should be different on Mac/Linux and Windows:
[<Literal>]
let connectionString = @"Data Source=./db/northwindEF.db; Version = 3; Read Only=false; FailIfMissing=True;"
...works on Mac/Linux but Windows wants this:
[<Literal>]
let connectionString = @"Data Source=.\db\northwindEF.db; Version = 3; Read Only=false; FailIfMissing=True;"
So there should be logic to parse DataSource file path in the SQL-Provider.

pezipink:
closing this, the contents and more stuff have been moved into the Development_V1 branch

PR #128, opened 2015-03-14
Improving PostgreSQL type mappings

janno-p:
Some changes to give better support for strange PostgreSQL data types. I'll give some reasoning about the changes:

* Added new types from NpgsqlDbType enumeration, that should be added in future versions. Also removed values which will not appear in `data_type` column, because they're actually aliases to another type.
* Created mappings to fix `data_type` names which differ from NpgsqlDbType enumeration value.
* Created function to check existence of Npgsql type, to avoid using types which appear in enumeration, but are not yet implemented by Npgsql (like NpgsqlLine which gives exception if used).
* Since postgreSQL has only functions (no procedures), it seems overly complicated to force separation between functions and procedures as for other database systems. Using `Functions` for all sprocs in case of postgreSQL.
* Removed member `GetSprocReturnColumns` from `ISqlProvider` and instead added `ReturnColumns` to `SprocDefinition` type. In such way I can extract sproc details in single place and it doesn't require adding driver specific details to common types, in case return columns member requires extra information about certain procedure.

Also some minor fixes in postgreSQL provider. Fixed compilation warning. Added some Linq methods to support `headOrDefault` and `exactlyOneOrDefault` inside query expression.

Populating individuals had issue when creating "As ..." types when values contain NULL values. Currently fixed by showing "&lt;null&gt;" in such case. Or would it be better to filter out null values (to favor entities which actually have real value)?.


pezipink:
Great work Janno - I will try and test this over this week.  @colinbull can you make sure your Oracle dbs work with this lot (sprocs, functions).

It looks like "Single" has been removed?  Why?  (exactlyOne)

colinbull:
Yep no problem. Should get to this late today or tomorrow.

janno-p:
> It looks like "Single" has been removed? Why? (exactlyOne)

Should be there. It just got shifted down because of FirstOrDefault.

pezipink:
ahh, ok, the diff is decieving

janno-p:
Also, I'd like to mention, that this PR only addresses type detection issues with PostgreSQL. The problem with Npgsql not being able to choose correct NpgsqlDbType with shared CLR types still remains, but it was broken before anyway. The issue might become more urgent though, because after this PR gets merged, more problematic types become available (mostly timestamp/time related types).

pezipink:
Yes, understood, we will address the type mapping thing separately

colinbull:
This looks good my end... Indeed we do still have the problem with the specific Npsql types, but that is true in Oracle and SqlServer as well with things like Geo or Oracle XML types. So I have merged. 

Sorry it took so long.

boeremak:
Thanks! Will test it out this weekend if I have the time.

PR #131, opened 2015-03-27
Bugs in the LINQ expression converter

netogallo:
The two new tests included in this pull request fail w/o the changes:

#1 It is not clear from the documentation (https://msdn.microsoft.com/en-us/library/bb549042%28v=vs.110%29.aspx) if Expression.New can be called when the "members" argument is null but apparently doing so causes the employeesFirstName1 function to crash.

#2 The F# compiler internally adds some calls to the FSharpFunc.Convert method which takes as an argument a Delegate from the Converter class. When constructed lambdas that are passed as arguments to this function, one must make sure they are a subclass of the Converter class. Adding this conversion makes the test employeesFirstName2 no longer crash.


colinbull:
Merged.. Thanks..

PR #143, opened 2015-07-20
Case-sensitivity fixed.

Thorium:
The problem was these (PostgreSQL-changes, in src/SQLProvider/Utils.fs) to common code that it stored table name, column names and sporc names to .ToUpper() and the corresponding case conversion was not used when fetching the items by different providers (GetTables, GetColumns): case sensitive databases couldn't find tables corresponding tables from the schema and corresponding columns from the tables.

There may be a good reason to modify the schema and table names, so "case-sensitivity manually changed" is now one parameter to the GetTables. I don't see any reason to modify the column and sproc names to uppercase as they are read from the system data and used as they are. So I removed the upper case from buildFieldName method, but if that is really needed, then the GetColumns should also transfer this information to the database level.

That was the major change. Other minor modifications of these two commits:
 - There is now one more optional parameter to the constructor: do you want to cast table names to upper or lower case or leave them as original. Default is ToUpper, as it was the current functionality of this repo.
 - sp_MSforeachtable is the correct system procedure case in SQL-server
 - Added database creation script for MySQL/MariaDB for easier setup
 - Added resolution path display to the error messages that the database driver file was missing from a path.
 -  Corrected MySQL delete procedure parameter name
 - For Mono, if you are ever planning to run CI test with the SQLite, the connection string that tells the database should understand unix-path-delimiter (slash, not backslash). This is now fixed.
 - Some unit test had absolute paths to creator's hard drive, changed to `__SOURCE_DIRECTORY__`
 - Added Antarcitca-test to ensure that the modifications are committed to the database.

I have played with these: MS-SQL, SQLite, MySQL (on Windows and on Mono/Mac).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.suo
 *.user
 *.sln.docstates
+*.vs/
 
 # Build results
 
--- a/src/DatabaseScripts/MSSQLServer/DATA.sql
+++ b/src/DatabaseScripts/MSSQLServer/DATA.sql
@@ -1,7 +1,7 @@
-EXEC sp_msforeachtable "ALTER TABLE ? NOCHECK CONSTRAINT all"
+EXEC sp_MSforeachtable "ALTER TABLE ? NOCHECK CONSTRAINT all"
 
 -- delete data in all tables
-EXEC sp_MSForEachTable "DELETE FROM ?"
+EXEC sp_MSforeachtable "DELETE FROM ?"
 Insert into dbo.REGIONS (REGION_ID,REGION_NAME) values (1,'Europe');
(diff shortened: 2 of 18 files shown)


pezipink:
great work, thanks very much! :) I will leave @colinbull to comment on the names / renaming of things as it was originally his work renaming stuff in various places.  I am happy for this to be merged ready for our legendary coming v1

Thorium:
Yes this won't change the names, the names are still like `context.''[HR].[CONTACTINFO]''.` but this just gives tools to manage situations  without total query failure when you have different case like CamelCase database table names, ``ContactInfo``.

Colin's approach is much cooler, `context.HR.Contactinfo.`, but it's not ready yet as it can't handle e.g. camelcase-names with case-sensitive database. But as Colin said, this should be merged to Colin's branch also.

PR #168, opened 2015-11-16
Fix incorrect detection of array type in "IN" operator generation

rotmoset:
To detect arrays being passed to the IN and NOT IN operators, a runtime type check was performed against "string array", however what was probably changed in this [changeset](https://github.com/fsprojects/SQLProvider/commit/35df630919a662fe368b0768e1d93d78b4059a40) the array actually being passed is of type "obj array". This PR fixes this check for all supported databases (this has actually been fixed multiple times before for individual vendors, see [this](https://github.com/fsprojects/SQLProvider/commit/0f0e3a8a8a26567927b636d634e66e62d3e28ecc) and [this](https://github.com/fsprojects/SQLProvider/commit/f7d379bc2010dd431e7eff69b33406d8f7e9732b), but this PR fixes this in uniform way).

### DIFF
--- a/src/SQLProvider/Providers.MSAccess.fs
+++ b/src/SQLProvider/Providers.MSAccess.fs
@@ -188,9 +188,9 @@ type internal MSAccessProvider() =
                         preds |> List.iteri( fun i (alias,col,operator,data) ->
                                 let extractData data = 
                                      match data with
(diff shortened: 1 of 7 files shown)


colinbull:
Nice... Thanks...

PR #169, opened 2015-11-17
Use OUTPUT clause instead of SCOPE_IDENTITY

mFingers:
I have a table with a uniqueidentifier as the primary key, along with a default constraint of NEWID().  I wanted that value back from an insert statement, but only IDENTITY columns are currently supported.   Using the [OUTPUT clause](https://msdn.microsoft.com/en-us/library/ms177564.aspx), I was able to get the new guid back.  Identity columns still work as expected, too.

This SQL is supported back to SQL 2005.

### DIFF
--- a/src/SQLProvider/Providers.MsSqlServer.fs
+++ b/src/SQLProvider/Providers.MsSqlServer.fs
@@ -580,9 +580,10 @@ type internal MSSqlServerProvider() =
                     |> Array.unzip
                 
                 sb.Clear() |> ignore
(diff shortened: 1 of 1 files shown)


colinbull:
Thanks..

pezipink:
Sorry, I was half of the fence this one as it will break < SQL 2005, I guess we can sort that out later though.

colinbull:
I was also on the fence but I'm sure anything < 2005 is out of support. TBH we should draw the line somewhere else we are just making a rod for our own backs in terms of support.

Issue #183, opened 2015-12-06
PostgreSQL 'Owner' and schemas


xenocons:
Default owner seems to be 'public', one can change this, but only to a single schema, making it hard to span multiple schemas. (perhaps this is bad DB design?).

Anyway, a temporary hackaround (sql injection) is to do this:
```
type ps = SqlDataProvider<
              ConnectionString = @"Server=sql;Port=5432;Database=XXXDatabase;user id=XXXX; password=XXXX; Pooling=false",
              DatabaseVendor = Common.DatabaseProviderTypes.POSTGRESQL,
              Owner = "agent' OR table_schema = 'analysis' or table_schema = 'reporting' or table_schema = 'storage''">
```

I suggest:
1. the type provider SqlDataProvider is updated to use a list for the 'owner' field
2. the type provider SqlDataProvider updates the help to mention it can be used for Postg, currently it says the owner field is Oracle Only (ill try find this and add a pull req).

Thorium:
Is this still a problem? This is kind of connected to #213

Thorium:
This should be ok now. If you are not happy with the current implementation, please reopen the issue and describe why.

JacobChang:
@Thorium I think the document should be updated as owner field is not Oracle Only anymore, I stuck with this today. If needed, I can add a PR on this

Thorium:
It would be nice!

PR #185, opened 2015-12-06
attempt to add support for owner lists.

xenocons:
in reference to point 2 on #183, this breaks the other type providers (Oracle and MySQL) due to inference. This will need a bit of extra effort to port similar changes to MySql\Oracle?

edit: added clarification about what is broken and what needs work.
might need to rename the parameter?


xenocons:
Delete this pull request, another one incomming.

PR #186, opened 2015-12-06
Update to support seperating owner with comma.

xenocons:
I have made this change in relation to tackling problems in #183. Note that the Oracle type provider also uses string array to hold the 'owner', so there may be an opportunity to make this more robust accross the multiple providers.

The reason to keep it as string is two fold, first, it is unweidly to provide a string list into the TP type constructor, I tried and failed. Second, this shouldn't break existing implementation for MySQL and Oracle (as I don't know what to do there).

One short coming of this is that it introduces inconsistency into the handling of 'owners', this needs peer review.

### DIFF
--- a/src/SQLProvider/Providers.Postgresql.fs
+++ b/src/SQLProvider/Providers.Postgresql.fs
@@ -10,7 +10,7 @@ open FSharp.Data.Sql.Common
 
 module PostgreSQL =
     let mutable resolutionPath = String.Empty
-    let mutable owner = "public"
+    let mutable owner = [| "public" |]
     let mutable referencedAssemblies = [| |]
 
     let assemblyNames = [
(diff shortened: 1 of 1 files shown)


colinbull:
Thanks for this, I'm not opposed to this. However we do already support it I believe by having multiple providers, e.g. 

      type Schema1 = SqlProvider<...., Owner="Schema1">
      type Schema2 = SqlProvider<....., Owner="Schema2"> 

 When I get a chance I'll merge this to the Development_V1 branch and have a look to see what needs to be done to make the providers more consistent.

xenocons:
Interesting, didn't think of the approach to have multiple providers, that is quite a good idea! If you do end up rejecting this PR on the owner splitting, please at the least make the Owner parameter get parameterised to avoid sql injection issues (it is unlikely that user input can taint the Owner field, but it could be an embarassing bug)

PR #205, opened 2016-01-06
Documentation formatting

### LINKED ISSUE #201: Documentation formatting
Currently the [main documentation page](http://fsprojects.github.io/SQLProvider/) looks pretty bizzare:

image

I think that maybe you need to leave an empty line before every list, and they will render correctly then (**that initial problem was fixed by #202**).

Another errors I've found:

- `Querying` link doesn't look like a link:
  
  image
- [Some of the core documentation pages](http://fsprojects.github.io/SQLProvider/core/general.html) differ from [the main page](http://fsprojects.github.io/SQLProvider/) in their design (just like they used an old template version)
- all the links in that old template navigation block are broken. Go to [General](http://fsprojects.github.io/SQLProvider/core/general.html) page and try to click any navigation link
- [Constraints & Relationships page](http://fsprojects.github.io/SQLProvider/core/constraints-relationships.html) looks very strange (as if it wasn't formatted at all); the same for [SQL Server](http://fsprojects.github.io/SQLProvider/core/mssql.html) and [SQLite](http://fsprojects.github.io/SQLProvider/core/sqlite.html)
- [CRUD page](http://fsprojects.github.io/SQLProvider/core/crud.html), [Programmability](http://fsprojects.github.io/SQLProvider/core/programmability.html), [Individuals](http://fsprojects.github.io/SQLProvider/core/individuals.html),[Composable](http://fsprojects.github.io/SQLProvider/core/composable.html) and [ODBC](http://fsprojects.github.io/SQLProvider/core/ODBC.html) pages are not found (returns 404) although I can see their sources in the project
- License link from the navigation block is not working

:warning: When I've tried to build the documentation using `./build.cmd GenerateHelp`, it threw an exception (see [here](https://github.com/fsprojects/SQLProvider/issues/201#issuecomment-169055068)). Maybe that's the core reason all these pages weren't generated or were generated with defects.

ForNeVeR:
This PR will close #201 completely. Two notable problems it fixes are:

1. FSharp.Formatting cannot process empty pages (and we still have some - e.g. `composable.fsx`, `crud.fsx`), so I had to fill them with some stub content.
2. `generate.fsx` was passing `"."` as the root directory to `Literate.ProcessDirectory`. That option was applied to every documentation page and caused troubles for pages inside of `core` subdirectory. I wrote a simple implementation of `getRelativePath` to solve this; now it passes proper relative paths.

   On the other hand, ProjectScaffold uses different strategy: it [defines a `root` variable](https://github.com/fsprojects/ProjectScaffold/blob/ab1dcfb302ad47b6bfd164197ad30a4503d76f2a/docs/tools/generate.template#L40-L46) pointing to file system or to web address depending on `RELEASE` preprocessor constant.

   I am not aware whether SQLProvider build process defines this `RELEASE` constant or not, and whether such change will have an impact on the documentation publication process, so I've decided to use safer option: implement my little `getRelativePath` and don't mess with the preprocessor. After all, if we can have one portable copy of documentation that can be deployed locally or to the website - why not have it?

### DIFF
--- a/docs/content/core/composable.fsx
+++ b/docs/content/core/composable.fsx
@@ -1,0 +1,4 @@
+(**
+# Composable Query
+Documentation stub.
+*)
--- a/docs/content/core/constraints-relationships.fsx
+++ b/docs/content/core/constraints-relationships.fsx
@@ -3,7 +3,7 @@
 (*** hide ***)
 #I "../../../bin"
 
-(*
+(**
 
 # Constraints & Relationships
 
--- a/docs/content/core/crud.fsx
+++ b/docs/content/core/crud.fsx
@@ -1,0 +1,4 @@
+(**
+# CRUD
+Documentation stub.
+*)
--- a/docs/content/core/general.fsx
+++ b/docs/content/core/general.fsx
@@ -23,12 +23,13 @@ connect to a variety of database sources in the IDE and explore them in a
 type-safe manner, without the inconvenience of a code-generation step.
 
 SQL Provider supports the following database types:
-    * [`mssql.html` file] MSSQL
-    * [`oracle.html` file] Oracle
-    * [`sqlite.html` file] SQLite
-    * [`postgresql.html` file] PostgreSQL
-    * [`mysql.html` file] MySQL
-    * [`odbc.html` file] ODBC (_Experimental_, only supports SELECT & MAKE)
+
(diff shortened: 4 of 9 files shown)


colinbull:
Thanks, I'll rebuild the docs today

PR #207, opened 2016-01-07
Fixed casing of the project page url

TWith2Sugars:
Old - Doesn't work: http://fsprojects.github.io/SqlProvider
New - Does work: http://fsprojects.github.io/SQLProvider


colinbull:
Thanks.

PR #210, opened 2016-01-14
format postgre identity name with double qoute

garfeildma:
PostgreSQL's identifiers (including column names, table names) are case
sensitive

http://www.postgresql.org/docs/current/interactive/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS


pezipink:
Thanks for this

Issue #188, opened 2015-12-06
PostgreSQL Npgsql Versioning


xenocons:
It appears that `NpgsqlTimeStamp` does not exist in the nuget version of Npgsql 3.0.4. This means tables containing columns that use `timestamp` will drop columns. This is due to a combination of the behaviour of line https://github.com/fsprojects/SQLProvider/blob/master/src/SQLProvider/Providers.Postgresql.fs#L148 which filters out the type 'timestamp' (as it does not exist) and https://github.com/fsprojects/SQLProvider/blob/master/src/SQLProvider/Providers.Postgresql.fs#L423 which surpresses unknown types.

I suggest we either lock down SQLProvider to a specific Npgsql, or make the parsing of this more lenient.

Thorium:
Is this fixed?

Thorium:
If this is still issue, we can do
```fsharp
"timestamp without time zone", (if isLegacyVersion.Value
                                then namemap "NpgsqlTimeStamp"      ["Timestamp"]
                                else typemap<DateTime>              ["Timestamp"])
```

Thorium:
Closed as obsolete tooling/runtime era.

PR #218, opened 2016-02-09
fix wrong typemapping for timestamp types in #188

coolya:
Fixes problems with npgsql 3.o which caused columns to disappear.


coolya:
Seems like it does not fix the problem fully. Looks like timestamps now only include the date not the time.

coolya:
Even without my change and the old npgsql 2.2.7 i have the same issue that `timestamp` field only contain the date part and no time.

coolya:
After doing the changes from #219 it worked fine with npgsql 3.0

colinbull:
Sorry, for the delay. Does this retain compatibility with 2.2.7, was it simply an incorrect type mapping?

colinbull:
Thanks.

coolya:
I tried it with 2.2.7 and it seems to work fine. Not sure if it was in general wrong. I wasn't able to find the docs for npgsql. There it seems to have worked with the old mapping but these classes where removed in 3.0.

PR #219, opened 2016-02-09
Fix DateTime columns would loose the time part due to typemapping.

coolya:
I'm not sure if this the correct way to fix it. At least for me it worked out quite well.

Since I have no idea why this type adjustment is present in the code it's hard to tell what it actually does. I think a comment at that line why it is present would put some light on the reader of the code.


colinbull:
Thanks.

PR #222, opened 2016-02-21
multi threading improvements

coolya:
I added some caching code that caches the parsed app.config if the type provider is in runtime mode. In my application that heavily uses actors this gave me some startup improvements and fixed a problem that monos `Path.GetTempFileName()` would create the same filename if two calls are almost at the same time.

In addition I added some safe guards around the `GetColumns` and `GetRelationships` because I was getting duplicate key exceptions there when called concurrently.


pezipink:
nice, thanks for this!

PR #231, opened 2016-03-20
Fix GetColumn Missing Type Handling.

xenocons:
Prevent GetColumn from silently dropping unsupported column types.


pezipink:
THANKS!

PR #234, opened 2016-03-30
Update QueryTests.fs

xenocons:
ensure we test for exists, which is currently not implemented, to help track unimplemented linq


colinbull:
Thanks.

Issue #179, opened 2015-11-30
Mapping over provided database types


forki:
@tforkmann and me are trying to convert from generated types to Domain model types:

    let private allElements = db.``[dbo].[C_Elements]``
    
    let mapToDomainModel element = 
      { X = element.X
        Y = element.Y
         ....
       }

    let AllElements = allElements |> Seq.map mapToDomainModel 

How can we tell mapToDomainModel that the input elements are of the generated type and the output elements are of `OurDomain.Element` type?

colinbull:
There is an example of this https://github.com/fsprojects/SQLProvider/blob/Development_V1/tests/SqlProvider.Tests/scripts/PostgreSQLTests.fsx#L97. Which I think is available on the master branch. I'd just note there are a lot of bug fixes on the Development_V1 branch we are trying to get a nuget out and close the last few issues.

forki:
cool. what is this generated mapping doing exactly?
Mapping by name and type?

colinbull:
It is a very niave mapping by name. Really thats all we have at runtime since the types are erased. However there is an overload to the `MapTo` function that allows you to define your own mappings see here https://github.com/fsprojects/SQLProvider/blob/Development_V1/src/SQLProvider/SqlRuntime.Common.fs#L188

pezipink:
If you need to write your own mappings you can get at the generated types to annotate them via the type name followed by the `dataContext` and then your mangled type name. eg

`CSMDB.dataContext.[api].[Benchmark]Entity`

forki:
A mapTo with a lambda is pretty nice

forki:
Yes what I'm looking for is the name of the type so that I can use it in the signature of my mapper function

pezipink:
yes see my comment above, you can get to them.

tforkmann:
@pezipink it worked! Thanks a lot!

PR #253, opened 2016-04-25
Documentation

tforkmann:
Babysteps in Documentation:

- Adding Docs for MSSQL.fsx
- Adding mappers.fsx

### DIFF
--- a/docs/content/core/mappers.fsx
+++ b/docs/content/core/mappers.fsx
@@ -23,7 +23,7 @@ This mapper will get sure that you always sync your types with types you receive
 *)
 
 type Employee = {
-    EmployeeId : int32
+    EmployeeId : int64
     FirstName : string
     LastName : string
     HireDate : DateTime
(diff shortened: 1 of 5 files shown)


pezipink:
Hi, thanks for this - I assume you are not aware of our `MapTo<T>` function?  It does that mapping to record types automatically for you, though it is a little slower

`MapTo<Employee>()`

Have you seen it?

tforkmann:
Ohh, sweet!
Gonna test it today.

tforkmann:
Works like a charm! Going to fix the docu

tforkmann:
Guess I got a little bit too excited there.
Its compiling but not working in my project. 

Suave returns a System.Reflection.TargetParameterCountException at the end.

We had that issue before that's why we created our own mapper.
https://github.com/fsprojects/SQLProvider/issues/179

We are only looking for the name of the type.
I tried property matching with 'MapTo' but I didn't really got that working. 
https://github.com/fsprojects/SQLProvider/blob/master/tests/SqlProvider.Tests/scripts/PostgreSQLTests.fsx#L144-L151

Guess I have to stick to the old mapper...

Issue #163, opened 2015-10-03
Possibility to create async queries?


Thorium:
Could there be a way to extend provider so that it would release the business logics thread while the database is doing its job?

![](http://i.imgur.com/DBPLRlP.png)

I don't want design-time async, just runtime.
Currently at least MSSQL and ODBC are supporting async queries.

One example, current code `cmd.ExecuteNonQuery() |> ignore` to something like `let! ok = cmd.ExecuteNonQueryAsync() |> Async.AwaitTask`

   1. I assume that there is no common async communication interface for all the providers. So non-async should be faked just with Task.FromResult. But creating extra tasks will have small hit to performance.
   2. What would be nice implementation? E.g: Would we like to add a new provider type like DatabaseProviderTypes.MSSQLSERVER_ASYNC or should we just extend the interfaces like FSharp.Data.Sql.Common ISqlProvider ?

How this would look good with async execution?

    let mattisOrderDetails =
        query { for c in ctx.``[main].[Customers]`` do
            // you can directly enumerate relationships with no join information
            for o in c.FK_Orders_0_0 do
            // or you can explicitly join on the fields you choose
            join od in ctx.``[main].[OrderDetails]`` on (o.OrderID = od.OrderID)
            //  the (!!) operator will perform an outer join on a relationship
            for prod in (!!) od.FK_OrderDetails_0_0 do 
            // nullable columns can be represented as option types. The following generates IS NOT NULL
            where c.CompanyName.IsSome                
            // standard operators will work as expected; the following shows the like operator and IN operator
            where (c.ContactName =% ("Matti%") && o.ShipCountry |=| [|"Finland";"England"|] )
            sortBy o.ShipName
            // arbitrarily complex projections are supported
            select (c.ContactName,o.ShipAddress,o.ShipCountry,prod.ProductName,prod.UnitPrice) } 
       |> Seq.toArray

C# has this `.ToListAsync()`, but should we have separate async query builder class or something?

I don't have immediate need for this so this is just an idea for a new feature.

Thorium:
I was investigating this a bit more.
There is no `Async` possibility for every provider.
Basically what you want to do is this:
```
async {
   use connection = ...
   do! connection.OpenAsync() |> Async.AwaitTask
   use command = new SqlCommand(query, connection)
   //parameters...
   let! something = command.ExecuteAsync() |> Async.AwaitTask
}
```
[...]

PR #257, opened 2016-05-01
Ability to make asynchronous database operations. #163

Thorium:
#163 
Ability to make asynchronous database operations, for all the providers. 

SQL-Query operations for lists (`open FSharp.Data.Sql`):

            query {
                for x in dbContext.Mytable do
                where (x.Something.IsNone)
                select (x)
            } |> Seq.executeQueryAsync

Crud operations to update, insert & delete:
`dbContext.SubmitUpdatesAsync()`

The old sync-methods and ways are still in place, so you can still choose those if you want to use your web/BL-server CPU-resources for heating.

These new async-ones will return F# `Async` class. You have to deal with that. My suggestion is that you say `Async.AwaitTask` and return that out from your web-server (like SignalR / Owin / WebApi / whatever).

I recommend NOT to do Async.AsParallel (as 3rd party data-connection-providers are usually not very good in parallelism) and NOT to use Async.RunSynchronously (as then you loose the benefits of async, that method is for losers).

Some testing done with MySQL and MsSql.

### DIFF
--- a/docs/content/core/crud.fsx
+++ b/docs/content/core/crud.fsx
@@ -64,3 +64,9 @@ row.Delete()
 Submit updates to the database
 *)
 ctx.SubmitUpdates()
+
+(** 
+
+Support also async operations: ctx.SubmitUpdatesAsync() |> Async.StartAsTask
+
+*)
--- a/docs/content/core/general.fsx
+++ b/docs/content/core/general.fsx
@@ -134,6 +134,20 @@ let customersQuery =
     |> Seq.toArray
 
 (**
+
+Support also async queries
+
+ *)
+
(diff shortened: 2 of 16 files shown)


tforkmann:
Sweet stuff! I might be able to test that with suave.

Thorium:
There is not so much changes as it seems:
`createInsertCommand` `createDeleteCommand` and `createUpdateCommand` are moved from each `member this.ProcessUpdates` to the top of each provider as the methods are reused in the `member this.ProcessUpdatesAsync`.

How could we get Mono-compilation work when `TransactionScopeAsyncFlowOption.Enabled` is definitely needed for transaction safety?

PR #258, opened 2016-05-05
Fix generated SQL when query expression doesn't contain any entity projections.

janno-p:
Found in PostgreSQL test script following failing query:

```F#
let employeesFirstNameNoProj = 
    query {
        for emp in ctx.Public.Employees do
        select true
    } |> Seq.toList
```

It results in SQL query starting with `SELECT FROM ...` (select part doesn't have content). This PR adds some default content to select part, so that simple query wouldn't fail: `SELECT 1 FROM ...`. It should handle simple cases where select has constant values which apply to all results. I don't know if it's possible to compose query expression which has no entity projections, but constant values vary between different rows. If it's possible, then current simple fix requires further improvements, because otherwise some operators (like distinct) would give unexpected results.

### DIFF
--- a/src/SQLProvider/Providers.MSAccess.fs
+++ b/src/SQLProvider/Providers.MSAccess.fs
@@ -241,6 +241,7 @@ type internal MSAccessProvider() =
 
             // first build  the select statement, this is easy ...
             let columns =
+                if projectionColumns |> Seq.isEmpty then "1" else
                 String.Join(",",
                     [|for KeyValue(k,v) in projectionColumns do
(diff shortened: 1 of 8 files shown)


Thorium:
Seems ok, thanks!
Should be merged soon as may cause conflicts with other changes.

janno-p:
It's interesting, that postgres 9.5 actually allows empty select clause (as I just found out), but older version complained about it. But since that's specific to certain postgres versions, the fix should be in order regardless.

Issue #246, opened 2016-04-18
Error inserting records with Postgres JSONB columns


ninjarobot:
### Description

Using the SQL provider to create a row with JSONB type fields does not seem to infer the correct type, and instead returns an error and no data is inserted:

> Npgsql.NpgsqlException : 42804: column "a3" is of type jsonb but expression is of type text

The type provider infers the field to be a string, so I am passing a string.  I'm not sure if there is any way to pass anything else here.  If I use the driver directly, it works successfully like this:

```
cmd.Parameters.AddWithValue("a3", NpgsqlDbType.Jsonb, "{ 'myval' : 3 }")
```

### Repro steps

Please provide the steps required to reproduce the problem

1. Create a table in a postgres database with a JSONB field.  This is the column declaration I am using:
`"a3" JSONB`

2. Open a context, create a record for a new row, and set to a string of JSON:
`row.A3 <- Some ("{ 'myval' : 3 }")`

3. Submit updates with `ctx.SubmitUpdates()`

### Expected behavior

JSONB data is inserted.

### Actual behavior

Driver raises this exception: Npgsql.NpgsqlException : 42804: column "a3" is of type jsonb but expression is of type text

### Known workarounds

None

### Related information 

* Operating system: OS X El Capitan
* Branch: Not sure what that means.  I'm using these packages, though:
```
  <package id="Npgsql" version="3.0.5" targetFramework="net45" />
  <package id="SQLProvider" version="1.0.0" targetFramework="net45" />
```
* .NET Runtime, CoreCLR or Mono Version: Mono 4.2.3
* Performance information, links to performance testing scripts: Not a performance issue

pezipink:
Hi

The postgres implementation does not currently support jsonb.  It was be a simple matter of telling it to use NpgsqlDbType.Jsonb instead of string in this case, or it may require some more work.

I also do not know if jsonb is supported in ngpsql 2.~ and if this would cause backwards compatibility issues (we have had other problems already with 3.~) .  

This should be a nice relatively easy one for someone to investigate, unfortunately I know little about postgres.

thoughts @colinbull  ?

Ross

colinbull:
I suspect the type mapping the JSONB is actually char[]? Looking at the (Npgsql docs) [http://www.npgsql.org/doc/types.html].

[...]

Issue #154, opened 2015-08-08
SQLProvider not working with Npgsql 3.0


psfblair:
Upgraded to Npgsql 3.0 and everything's gone wonky with postgres.

Intellisense shows me dataContext.``[PUBLIC].[FORTUNE]`` but won't let me put .Individuals after it, and basically won't autocomplete anything at that point.

If I do:
dataContext.``[PUBLIC].[FORTUNE]`` |> Seq.map (fun row -> row.id)

It says "the field, constructor, or member 'id' is undefined." Autocomplete gives me members of SqlEntity such as ColumnValues and DataContext.

janno-p:
I tried Npgsql 3.0 with current master and after fixing the issue with string types (#155), the test script loaded and everything the script expected was loaded correctly.

There seems to be an issue with stored functions at the runtime though. It looks that there are some changes regarding functions returning refcursor. Querying from such function returns `"<unnamed portal n>"` as a result. It seems Npgsql 2.x versions handle that result internally, but Npgsql 3.0 expects clients to handle the portals by themselves (or at least that portal is returned at the moment and it doesn't fit with record type constructor in `SqlEntity.MapTo` member method).

kunjee17:
I tried using Npgsql 3.0.3 and things break. Type provider started crying for mono.security.dll. I don't know why. For now I rolled back to 2.2.7. 

let me know if any other details are needed for issue.

colinbull:
Yeah, I had a look at this tonight and there is quiet a few things we need to change to support >= 3.0, so for the moment at least 2.2.7 will be the preferred postgres version. As I don't see an easy way of supporting both. Changing this to up-for-grabs ;)

kunjee17:
@janno-p hey, you have a branch where Npgsql 3.0.3 working? I am so wanted to pull that one.

janno-p:
Hey @kunjee17, my npgsql3.0 branch is actually the one where I tried using >= 3.0 version of Npgsql. I made some small changes as described in my previous comment, and these should be already merged into the main repository. So, nothing extra in there.

colinbull:
Indeed these changes are merged. In the limited things I have tried, it seems that all the query support is there. The only thing that doesn't work is the stored procedures.

kunjee17:
@janno-p and @colinbull thanks guys. I'll have a look again at my code with update. Will post the update. Thanks.

kunjee17:
[...]

Issue #125, opened 2015-03-08
Double precision in Postgres does not work anymore


boeremak:
When using float8 in Postgresql it gets converted into "Double precision" database type. This does not get properly converted by the type provider.

I've added the following line into the Postgres provider and recompiled:

yield { ProviderTypeName = Some "double precision"; ClrType = (typeof<double>).ToString(); DbType = DbType.Double; ProviderType = Some 22; }

This does seem to work. As "double precision" seems to be just a different name for float8 I can't really say if this breaks anything.

If you could look into it I would be much obliged.

pezipink:
@janno-p any thoughts on this? I know little to nothing about postgres !

janno-p:
I'm familiar with the issue. I'll put together some fixes for postgres types what I've discovered so far. There are some additional types with same issue, and at the same time some mappings are not needed.

boeremak:
Are we constrained by the types that Ado.Net provides? There are some more types that would be useful in the provider that Postgres provides.

pezipink:
Yes, this a purely ADO.NET based system, we are constrained by whatever the chosen ADO.NET provider - in this case ngpsql - is able to give us.

boeremak:
Ok, makes sense. Thanks

Issue #123, opened 2015-02-26
Generated types should hold provider specific data types for columns


janno-p:
While trying to improve PostgreSQL type handling, I created some tests for trying out various type combinations and aliases available in PostgreSQL. But it doesn't work at all, since generated types don't hold provider specific type information and at runtime type detection isn't enough (since multiple provider types can map to single clr type).

For example, in postgre I have 4 types I could map to DateTime (time, timestamp: both with or without timezone info). They all should map to System.DateTime, but at the runtime the System.DateTime type picks always same provider type (currently NpgsqlDbType.Time, since it's first in the mappings). It's possible to change order of mappings as in PR#117, but it is not suitable for all types (like string, which can be various other types in addition to text).

It seems that SqlEntity should hold the provider specific type enum value in addition to data, because then it's more certain, the value gets correct type in query-s.

But this probably means that more code that is currently general, should be moved to provider specific implementation, so it's probably reasonable to implement when specialized type providers (PR#85) gets finished and approved.

This is probably not postgresql specific problem, but more general issue?

janno-p:
Now that I come to think of it, I would be better of with using certain wrapper clr types for specific provider types. I will try out that approach first.

pezipink:
How does Npgsql deal with this issue?  Presumably it does a similar thing where it "remembers" something that is now a CLR DateTime should actually be a NpgsqlDbType.Time, or is the user expected to manage this?

janno-p:
Npgsql has internal mapping for various clr types (actually it seems they're making some changes in type system, future versions seem to implement typehandlers for all possible NpgsqlDbType enumeration values). But those mappings are overlapping for some enumeration values and system types.

Npgsql could probably handle types well most of the time, but current implementation might have some issues. If I remember correctly, then one issue is with `PostgreSQL.createCommandParameter` function. It first sets the value of the parameter and after that assigns `DbType` enumeration value. But in case of `ProcessUpdates` call, the type info is usually not given and defaults to String type. It's possible that most types handle it well, but I'll have to check (I don't remember what types exactly caused problems for me, if they're Npgsql specific geometric types, then it's not that much of a problem).

To recover I tryed some workarounds. One thing that works for `ProcessUpdates`, is to look up column info cached in provider and pass the mapping to parameter creation. That way the provider uses always same type it got from database (presuming the mappings are correct :)).

[...]

PR #261, opened 2016-05-06
Npgsql 3 stored procedures and improved PostgreSQL type support.

janno-p:
Rather large PR which should fix most of important issues that concern PostgresProvider:
* Issue #246
* Issue #154
* Issue #125
* Issue #123

## Changes
* Fixed stored procedures which return refcursor type with Npgsql 3 driver.
* Added support for stored procedures which return table of style results which are preferred over refcursor (both Npgsql 2 and 3)
* Reviewed Npgsql 2 and 3 supported types and corrected mappings accordingly.
* Added tests to PostgreSQL script which confirms that all types are usable and saved/loaded correctly.
* Added column info to SqlEntity type which is required by PostgresProvider to make sure that correct NpgsqlDbType values are used while executing commands.

## Notes
* Changes were made so that both (2 and 3) version of Npgsql would work, although whenever possible, the later version should be preferred, because older driver supports fewer types and some of them work in limited mode or not at all. Should consider dropping Npgsql2 support entirely.
* Most common types resolve to common .NET types like DateTime, TimeSpan, int, string etc., but more specific ones use types from Npgsql assembly. So when schema contains those special types (NpgsqlBox, NpgsqlTsQuery, ...) and they're used, Npgsql assembly should be referenced from the calling script or application.


Thorium:
Looks good to me!
I'll take more like @forki's approach to the release process:
Release now and when there are problems let's fix it later.

Thorium:
By the way, reference Npgsql is pinned to 2.2.7
Is this on purpose?

janno-p:
It's not required anymore, both should work. Npgsql 3 is better though, so I would recommend to unpin it. But changing to Npgsql 3 will break the postgresql test script, so I can do it myself and fix the script aswell.

PR #270, opened 2016-05-17
adding documentation for composable queries

tforkmann:
Sorry for the delay. 
Here the the documentation for the composable queries

### DIFF
--- a/docs/content/core/composable.fsx
+++ b/docs/content/core/composable.fsx
@@ -1,4 +1,61 @@
+(*** hide ***)
+#I @"../../files/sqlite"
+(*** hide ***)
+#I "../../../bin"
+(*** hide ***)
+[<Literal>]
+let connectionString = "Data Source=" + __SOURCE_DIRECTORY__ + @"\northwindEF.db;Version=3"
+(*** hide ***)
+[<Literal>]
+let resolutionPath = __SOURCE_DIRECTORY__ + @"..\..\..\files\sqlite"
+#r "FSharp.Data.SqlProvider.dll"
+open FSharp.Data.Sql
+
 (**
-# Composable Query
-Documentation stub.
+# Composable Queries
+
+The SQLProvider supports composable queries.
+With composable queries you can combine two queries in multiple ways, and one query can be used as the building block for the other query.
+To see how this works, let’s look at a simple query:
+*)
+
+let query1 =
+    query {
+      for customers  in ctx.Main.Customers do
+      where (customers.ContactTitle = "USA")
+      select (customers)} |> Seq.toArray
+
+(**
+The variable that is returned from the query is sometimes called a computation.
+If you write a foreach loop and display the address field from the customers returned by this computation, you see the following output:
+* GREAL|Great Lakes Food Market|Howard Snyder|Marketing Manager|2732 Baker Blvd.|Eugene|OR|97403|USA|(503) 555-7555|
+* HUNGC|Hungry Coyote Import Store|Yoshi Latimer|Sales Representative|City Center Plaza 516 Main St.|Elgin|OR|97827|USA|(503) 555-6874|(503) 555-2376
+* ...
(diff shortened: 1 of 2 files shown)


Thorium:
I thought `|> Seq.toArray` will execute the first query to the database and composable query means that you can do logics to compose just one database-SQL-query from multiple queryables...?

Thorium:
Something like this:
```
open System.Linq

/// First, let's define a composable query outside the main query
let companyNameFilter inUse queryable = 
	let queryable:(IQueryable<CustomersEntity> -> IQueryable<SqlEntity>) =
		match inUse with
		|true -> (fun iq -> iq.Where(fun (c:CustomersEntity) -> c.CustomersEntity = "The Big Cheese"))
		|false -> (fun iq -> iq.Where(fun (c:CustomersEntity) -> c))
	queryable

/// Then the main query
let query1 =  
    query {  
        for customers  in ctx.Main.Customers do  
        where (customers.ContactTitle = "USA")  
        select (customers)}
 
/// and now call: 
let query2 =  
    companyNameFilter true query1 |> Seq.toArray
```
ping @pezipink: this is also benefit of using LINQ Queryable, we can use .NET LINQ functions for IQueryable, outside the `query { ... }`.

tforkmann:
OK then I didn't understand composable correcly. I will fix the docs then. Sorry for that.

pezipink:
Guys, the "composable query support" means the integration with this library http://fsprojects.github.io/FSharp.Linq.ComposableQuery/

Thorium:
I think that we are not in the same problem space what they are having with usual O/R-mapping like LINQ-to-SQL and EntityFramework. These have the problem of entity context having so many roles:

1. It manages the "domain model" if not-in-POCO. And even if you have POCO, you still have to manage the mapping by rules of that framework.
2. It manages the connections to database
3. It manages the generation of database querys

You want the database transactions be short and the connection open a minimum time. So with EF/LINQ you want to inject your query logics outside of the DB-context to the query. But the problem is that is so hard to get the query done when you don't have your entity-objects ready.

Now, with SQLProvider:

1. User don't just care about the objects or mapping. Just write your business logics.
2. Connections are done with `TypeProviderConnection.GetDataContext()`, but as far as I know, it makes a connection per query.
3. Queries are managed with `query { ... }`syntax. There is not so much need for executing lambdas outside the context.

But! Our front page says we support composable query, and I don't see it yet. Let's take an example:

```
    let qry1 =
        query { for u in dbContext.Users do
                select (u.Id, u.Name, u.Email)
        } 

    let qry2 =
        query { for c in dbContext.Cars do
                select (c.UserId, c.Brand, c.Year)
        } 
                        
    query { for (i,n,e) in qry1 do
            join (u,b,y) in qry2 on (i = u)
            where (y > 2015)
            select (i,n,e,u,b,y)
        } |> Seq.toArray
```

I would expect that this generates one SQL clause having the join and where, and sends the small resultset from my Australian database server to my UK business logics server (as you know, network is typically the bottle-neck). Instead it will transfer all the users from Australia to UK, then transfer all the cars from Australia to UK, then throws away most of the results while executing the logics in UK.

I don't know about this FSharp.Linq.ComposableQuery library, it seems not to be very active?

tforkmann:
Ok guys,

I guess for creating the right documentation we need to discuss the "feature" composable a little bit more in detail.
Should I close the PR and create an issue instead?

I could also document all usecases of composable queries. People (like me) might want to know that those features exist and are possible to use with the SQLProvider....

Thorium:
Even when this documentation is bit wrong subject / off-topic, I would accept the PR as it's better than "Documentation stub."

tforkmann:
Ok i will just add your two composable examples and I will mention that the composable support means the integration with the FSharp.Linq.Composable library. I will update the PR quickly...

tforkmann:
Ok guys. Hope this brings us in the right direction. Added IQueryable and FSharp.Linq.Composable example and an general definition of Composable queries.

Thorium:
Thanks!
Merged. I made some minor corrections.
http://fsprojects.github.io/SQLProvider/core/composable.html

Btw, if you are interested in more details of LINQ from C# point of view (which the document slightly had) read this through and give me some feedback what to improve: https://github.com/Basware/LINQ-Tutorial

Issue #256, opened 2016-04-30
"System.Byte[]" is inserted as the value of BLOB (MySQL)


vain0:
### Repro steps
1. Create a table containing a field of BLOB type. (MySQL)
2. Insert a row into the table with SQLProvider.

This is an example.

```fsharp
open System.IO
open FSharp.Data.Sql

type sql =
  SqlDataProvider<
      Common.DatabaseProviderTypes.MYSQL
    , ConnectionString  = "Data Source=localhost; Port=3306; user=root; password=;"
    , ResolutionPath    = @"C:/Program Files (x86)/MySQL/MySQL Connector Net 6.9.8/Assemblies/v4.5"
    , UseOptionTypes    = true
    , Owner             = "test"
    >

[<EntryPoint>]
let main argv =
  let ctx = sql.GetDataContext()
  let db = ctx.Test
  let item = db.items.Create()
  item.Content <- Array.create 100 (byte 'a')
  ctx.SubmitUpdates()
  0
```

### Expected behavior

The assigned binary data (byte[]) should be inserted as the value of the field.

### Actual behavior

"System.Byte[]" (13byte) is inserted.

### Related information 
* Windows 10
* SQLProvider 1.0 (from NuGet)
* F#4.0
* .NET framework 4.5

Thorium:
Tried and yes this still exists.

Thorium:
Hmm, debugged a bit: It creates correct SQL clause `INSERT INTO test.Test (items) VALUES (@param0); SELECT LAST_INSERT_ID();` into `cmd:IDBCommand`, and adds the binary array to a parameter with 
`values |> Array.iter (cmd.Parameters.Add >> ignore)` when it's still correct binary. 

So this could be MySql.Data.dll problem with .Parameters.Add.

vain0:
Thanks for your work.

Now I should try to insert a blob data by using only MySqlData.dll & Parameters. My script is the following, which succeeded to insert actual value of blob type.

```fsharp
open System.Configuration
open System.Data
open MySql.Data.MySqlClient

let connectionString =
  let b =
    MySqlConnectionStringBuilder
      ( UserID = "root"
      , Password = ""
      , Server = "localhost"
      , Database = "test"
      )
  in b.ConnectionString

let withDb f =
  use connection = new MySqlConnection(connectionString)
  try
    connection.Open()
    f connection
  finally
    connection.Close()

let insert connection =
  let value = Array.create 100 (byte 'a')
[...]

PR #278, opened 2016-06-10
Enhancing the SQLite provider to correctly type parameters

JohnDoeKyrgyz:
I was not able to save byte[] values to a SQLLite BLOB column. This change fixes that by making sure that all DbParameter objects used in the SQLite provider have the correct DBType. 

It looks like issue #256. is basically the same issue for the MySql provider.


Thorium:
Just to inform, not to blame (the blame is the ugly ref-parameter):
It seems that this commit broke the SQLite param arrays (IN-clauses) to contain just one parameter (as all are called param1 and SQLite thinks they are the same). That is now fixed.

Issue #255, opened 2016-04-28
Oracle provider error: Column 'PACKAGE_NAME' doesn't belong to table Functions.


vilinski:
### Description

Subject error by using oracle type provider, probably due limited read rights

### Repro steps

1. Create an empty folder
2. Open it in VS Code
3. Paket init; Paket add NuGet SqlProvider
4. Create fsx file referencing the dll and defining some type with Oracle type provider
4. The oracle connection string for big instance with limited read/write rights

### Expected behavior

Please provide a description of the behavior you expect.

### Actual behavior

`SqlProvider.fsx(3,14): error FS3021: Unexpected exception from provided type 'FSharp.Data.Sql.SqlDataProvider,DatabaseVendor="4",ConnectionString="myconnectionstring",ResolutionPath="c:\\\\Some\\\\valid\\\\path\\\\to\\\\packages\\\\Oracle.ManagedDataAccess\\\\lib\\\\net40"+dataContext' member 'GetMethods': The type provider 'FSharp.Data.Sql.SqlTypeProvider' reported an error: Column 'PACKAGE_NAME' doesn't belong to table Functions.`

### Known workarounds

No known workarounds

### Related information 

* Operating system Windows 7 Enterprise
* .NET Runtime 4.5

mofus:
Has anyone found a workaround for this?

pezipink:
one for you @colinbull  ?

Thorium:
Quite optimistic for push to close the bug directly. :-)

Thorium:
Can you test this with new version please?

mofus:
I just tested this version and it works for me. Thanks for the great work!

Thorium:
Great!

vilinski:
Also tested and can confirm - the exception is away. Thank you!
Unfortunately it is still quite useless for me because of such crazy delays in the intellisence and compilation. The number of tables and columns in our corp server is just sick. Wonder whether there are some ways for improvement. May be table name filter like in "Oracle SQL Developer"

colinbull:
Humm, we do have a lazy loading strategy for SQL server, maybe the same
thing could be employed for Oracle.. I don't really remember how the oracle
side works thou.

pezipink:
In general the provider is supposed just load the names of the tables/views, then the columns of a table when you access it. What was killing us in the past was the sprocs, but I thought we fixed the lazyness of those? I don't remember how Oracle works either !

[...]

PR #287, opened 2016-06-20
Add TableNames static parameter

vilinski:
Hi,

the changes to cure the TP performance problems on very big Oracle instances, reported in the comments of https://github.com/fsprojects/SQLProvider/issues/255

The new 'TableNames' string parameter is a comma separated table names list. Allows "table_name_%" to handle them as in 'LIKE' query.

In short, I have replaced some 'getSchema' calls with pure sql queries.
Almost no latencies now.

Please review


Thorium:
Thanks! Will the default be all? So that the behaviour is the one it used to be if the parameter is not provided?

vilinski:
@Thorium Yes, default is unfiltered

PR #288, opened 2016-06-21
Three small documentation fixes

jackmott:
Escape the less than and greater than characters in parameters.fsx so Literal doesn't disappear
Fix the order of parameters given to SqlDataProvider for PostgreSQL and MSSQL


Thorium:
Great, thanks!

PR #295, opened 2016-06-22
Supported query expression documentation

jackmott:
This is intended to be a template we can use to track what query expression keywords are supported, with notes about limitations when they are only partially supported.  Once we have this filled out I can convert this to the fsx format that gets generated into the documentation website. For now it is easier to preview as markdown.


jackmott:
I've just filled out the few things I know to be supported, I'll need help filling this out completely.

Thorium:
Thanks, great!

I think besides metioned we support at least (and there is also unit test project for these):
exactlyOne
count
minBy
maxBy
sortBy
sortByDescending
thenBy
sumBy
averageBy
distinct
head
skip
take

I don't know about these nullable ones, may work or may not...
But I also know there is a problem with SQLServer on sortDescending+skip+take combination. On others it's ok.

Missing fromm this list:
if (supported)
let (not supported)

jackmott:
I've updated to reflect your notes

PR #298, opened 2016-06-23
WIP: Reimplement Oracle Packages and Procedures and functions + Perf improvements

colinbull:
This PR, re-implements oracle packages, procedures and functions. This is done in a completely lazy way, so there should be an improvement in intelli-sense response, On my test DB with ~300 packages and about ~1700 sprocs and functions in total there is a significant speed up everything should now be lazy and I no longer depend on `getSchema` to get the procedure parameters.
 
Additionally this PR also fixes a bug around handling oracle specific types that implement INullable. The type tree hasn't changed from the original implementation. So packages are still accessed through the `Package` type that sits at the same level as schema.

    let contract ref = 
         seq { 
            for cont in ctx.Packages.PgGmsContractUtils.PrGetContractByRef.Invoke(ref).P_CURSOR do
                 yield cont.ColumnValues |> Seq.toArray
          } |> Seq.toArray

This is a work in progress as I have only tested this with oracle at the minute. It shouldn't have affected any other of providers but just in case I gonna just test against my example databases. But if others could test aswell that would be good


pezipink:
nice!

colinbull:
Has anyone had a chance to test this?

Thorium:
I don't have aby Oracles near... Do you have perf increases to other providers as well? :-)

colinbull:
@pezipink Any chance you could give this a once over. If it is good I'll merge it.

Thorium:
You already merged master so this is not needed anymore.

Thorium:
Sorry my bad, GitHub was bit unclear. I tried to close my PR to your repo, not your PR. :-)

colinbull:
If no one has any issues I'm going to merge this.

Thorium:
Ok for me. Then you can focus on helping me with the group-by ;-)

The current problem is that entity executing query ReadEntities will return SqlEntities while I would like it to return Grouping<SqlEntities> and the grouping should have a key. Now, how do you read generic group-by response from all the dbs to something meaningful format... Also another problem is to get the aggregate columns: now after the group-by there will be another lambda, "select", which is the aggregation, and it should be serialized to the group-by-query. So maybe something like the join currently does.

Thorium:
Hmm. I merged this, but one question after merge... :-)
Why
`FSharp.Core\lib\portable-net45+netcore45+MonoAndroid1+MonoTouch1\FSharp.Core.dll` -> 
`FSharp.Core\lib\net40\FSharp.Core.dll`
...as this reference comes from Paket, it should be defined to paket.dependencies file and not modified directly.

PR #305, opened 2016-07-25
Fixing individuals link

julien-leclercq:
The individuals link is broken, pointing to http://fsprojects.github.io/SQLProvider/core/core/individuals.html instead of http://fsprojects.github.io/SQLProvider/core/individuals.html


Thorium:
Thanks!

Issue #310, opened 2016-08-08
Expose ContinueUpdateOnError on strongly typed data tables


isaacabraham:
### Description
It is sometimes necessary to perform an insert of records and continue inserting even if one record fails. With the raw DataAdapter you can set the ContinueUpdateOnError property, but this is not exposed by the strongly type data table model. Perhaps adding optional arguments on the Update() call would suffice?

Thorium:
I kind of like this idea...

Thorium:
What do you think of version in this branch thorium-exceptionhandling?

isaacabraham:
Comments on the PR

Thorium:
So basically wrong repo.
I'll close this for now.

PR #311, opened 2016-08-08
overload for SubmitUpdates

Thorium:
Possible solution for #310

### DIFF
--- a/src/SQLProvider/SqlDesignTime.fs
+++ b/src/SQLProvider/SqlDesignTime.fs
@@ -453,8 +453,21 @@ type SqlTypeProvider(config: TypeProviderConfig) as this =
                 yield! Seq.cast<MemberInfo> it
 
               yield! containers |> Seq.map(fun p ->  ProvidedProperty(p.Name.Replace("Container",""), p, GetterCode = fun args -> <@@ ((%%args.[0] : obj) :?> ISqlDataContext) @@>)) |> Seq.cast<MemberInfo>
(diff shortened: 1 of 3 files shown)


isaacabraham:
Confused - I was talking about the Update() method on the provided data table?

pezipink:
We only use DataTables internally,  are you looking at the right project? :)

pezipink:
In other news, I like the idea of doing something about this but I think it warrants further discussion.  

Allowing failed inserts in a batch invalidates the idea of a transaction completely so I would consider having this as something separate from our standard ORM style processing model.  Currently we always use transactions, don't we?  So what exactly is this change doing?  If one entity fails, the whole lot gets rolled back and the entity statuses are unchanged, right?   That is what it used to do as far as I recall?

isaacabraham:
Doh. This is for the wrong repo. Sorry.

isaacabraham:
At least someone else found this to be a good idea though!

Thorium:
hehe

Thorium:
If the whole transaction fails, then

- The default behaviour should be as is (e.g. for compatibility reasons).
- User could decide what to do with the failed objects: If they are left to the context then probably every future update will also fail. So probably we just want to cancel the whole transaction and clear all the objects and pretend that nothing has happened. Not always, but as I would think this is quite general functionality (e.g. with a webserver where one user managing to feed invalid objects shouldn't mess all others) we could at least give users easy option to do this.
- It would be also nice if the error message would tell the failed objects

Thorium:
We don't need more source code just for fun so if there is not actual need for this I close it now without merging.

Issue #294, opened 2016-06-22
special mysql column name, e.g. from


ingted:
### Description

insert update a mysql table with special word column name like 'from' will cause mysql command error

### Expected behavior

should be  `columnname`, not only columnname

### Known workarounds

insert:
("\`" + (String.Join("\`, \`",columnNames)) + "\`")

update:
(String.Join(",", data |> Array.map(fun (c,p) -> sprintf "\`%s\` = %s" c p.ParameterName ) ))

### Related information 

* Operating system
SERVER 2012 R2
* Branch
master
* .NET Runtime
4.6.1

Thorium:
As there is this workaround, are you going to make a PR or are you expecting someone else fix this? :-)

ingted:
Hi Tuomas,

I think you saw my code...
I will make a PR later in these days.^^

(Sorry for that I am not very familiar with Git mechanism, still in
learning.)

Regards
Anibal

Thorium:
Great! Just push to your GitHub-fork and then go to your fork web page and press the pull-request button. :-)

I assume the parameter count for insert/update is never zero. :-)

Br
Tuomas

Thorium:
Merged and in NuGet version 1.0.25.

PR #314, opened 2016-08-11
Postgres & MySQL add quotation mark to column names in generated SQL statement

ingted:
[ISSUE 294](https://github.com/fsprojects/SQLProvider/issues/294)

### DIFF
--- a/src/SQLProvider/Providers.MySql.fs
+++ b/src/SQLProvider/Providers.MySql.fs
@@ -271,7 +271,7 @@ type internal MySqlProvider(resolutionPath, owner, referencedAssemblies) as this
         sb.Clear() |> ignore
         ~~(sprintf "INSERT INTO %s (%s) VALUES (%s); SELECT LAST_INSERT_ID();"
             (entity.Table.FullName.Replace("[","`").Replace("]","`"))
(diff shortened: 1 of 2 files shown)


ingted:
I could build these changes in my computer... just don't understand why it fails here...

ingted:
It said "The command "sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF" failed and exited with 2 during ."

This should be the build server issues, not code issues...

Thorium:
Thanks...
For MySql there are both Replace "[" with quote, and the new addition of quote. So is there a danger of creating multiple quotes?

ingted:
Hi Thorium,

The code changes are what I use in our production environment ETL application.
I think there should not any  "[" or "]" in any MySQL code... so the replace just do nothing...

Sorry that I have to say I didn't think very deep about it (I just ignore...)
So IMO, my changes are not a good change for long term...

Would you mind give me some suggestions?
I will do another change and try making it more precise...

Thank you!!

Thorium:
My bad, the other one was table name and the other one was column names.
This is ok, thanks!

PR #315, opened 2016-08-14
Support for composite keys.

Thorium:
Initial idea is simple:

Change

```
let primaryKeyLookup = ConcurrentDictionary< tablename:string, primarykey:string >()
```

...to be...
```
type KeyColumn =
| Key of string
| CompositeKey of string list

let primaryKeyLookup = ConcurrentDictionary< tablename:string, primarykey:KeyColumn >()
```

and then when there is update or delete command, just match that if CompositeKey, parse AND combination:

```
DELETE FROM table WHERE myKey1 = @pk0 AND myKey2 = @pk1;
```

Tested to work with MSSQL-Server and Access.


Thorium:
Actually as one primary key is just a case of one in the list, the code could be cleaned to be just string list.

PR #325, opened 2016-08-26
Even better SQL generation!

Thorium:
More efficient select column selection. (e.g. not-used let-columns will be cut away)
Both select and where can have external variables.
I copied some boolean algebra from @mavnn


Thorium:
This is ready and working, but I would be happy for a little code review that I didn't make any mistakes on boolean logics. :-)

PR #333, opened 2016-09-18
Option to use System.Data.SQLite under Mono

wezeku:
**For SQLite**
I've added an option `SQLiteLibrary` to specify whether SQLProvider for SQLite should load System.Data.SQLite, Mono.Data.SQLite or make an automatic choice (the current version always chooses automatically).

Usage:

    type sql = SqlDataProvider< 
                    DatabaseVendor = Common.DatabaseProviderTypes.SQLITE,
                    SQLiteLibrary = Common.SQLiteLibrary.SystemDataSQLite >

The choices for `SQLiteLibrary` are `SystemDataSQLite`, `MonoDataSQLite` and `AutoSelect`. Default is `AutoSelect`, so that existing code doesn't break.

**Reasons for using System.Data.SQLite under Mono**
- If you build SQLite.Interop on the Mono machine, you get SQLite virtual table support (not present in Mono.Data.SQLite).
- In Mono.Data.SQLite, classes don't have the same capitalisation as System.Data.SQLite, so you can't use the same source code as you do in .NET and just leave both the Mono.Data.SQLite and the System.Data.SQLite dlls in your program directory. You'd have to resort either to \#defines and conditional compilation or to using Mono.Data.SQLite under both .NET and Mono.
- The System.Data.SQLite NuGet package is actively maintained, whereas Mono.Data.SQLite is rarely updated. As of September 2016, the former is about a week old, and the latter is older than 1.5 years. This means that bug fixes (e.g. DateTime conversion throwing exceptions when using time zone info) remain unreleased for Mono.Data.SQLite.

### DIFF
--- a/docs/content/core/sqlite.fsx
+++ b/docs/content/core/sqlite.fsx
@@ -40,13 +40,23 @@ the ResolutionPath parameter at all, but you still need to copy the x64 and x86 
 let resolutionPath = __SOURCE_DIRECTORY__ + @"/../../../tests/SqlProvider.Tests/libs"
 
 (**
+### SQLiteLibrary
+
+Specifies what SQLite library to use. This is an `SQLiteLibrary` enumeration, defined in the `FSharp.Data.Sql.Common`
+namespace, which has the following members:
+
+- **`AutoSelect`** - Uses System.Data.SQLite under .NET and Mono.Data.SQLite under Mono. This is the default.
+- **`MonoDataSQLite`** - Always uses Mono.Data.SQLite.
+- **`SystemDataSQLite`** - Always uses System.Data.SQLite.
+
 
 ## Example
 
 *)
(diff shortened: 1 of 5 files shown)


Thorium:
That's interesting. Could other providers also have this kind of need of changing the drivers? Like Oracle to Devart Oracle driver? Ping @colinbull

Thorium:
Could you update this description also to sqlite.fsx documentation? :-)

wezeku:
I've added documantation for the SQLiteLibrary parameter to sqlite.fsx. I'll write instructions on how to compile SQLite.Interop for Linux later, when I get time. I feel that's a bit out of scope for the SQLProvider SQLite docs, but it would be good if we could put it in a sub page and put a link to it in sqlite.fsx.

Thorium:
Thanks!

Issue #334, opened 2016-09-22
Documentation fix for PostgreSql


s952163:
### Description

In the [PostgreSQL doc](http://fsprojects.github.io/SQLProvider/core/postgresql.html) the documentation instantiates the type provider with:

### Repro steps

```
let sql =
    SqlDataProvider<
        dbVendor,
        connString,
        resPath,
        indivAmount,
        useOptTypes>
```
This should be:

`type sql= SqlDataProvider<...>`

PR #335, opened 2016-09-22
Doc fix in postgresql.fsx

s952163:
change `let sql` to `type sql` for issue: https://github.com/fsprojects/SQLProvider/issues/334


Thorium:
Thanks!

Issue #212, opened 2016-01-16
Will this provider generate types based on DB structure?


xperiandri:
Will this provider behave similar to Entity Framework and generate types for database structure?

colinbull:
Yes, see http://fsprojects.github.io/SQLProvider/for more info. The major difference is the types are erased not generated.

xperiandri:
I can't understand how to achieve what I want.
I want to return some data from DB from ASP.NET WebAPI. If I return items from table they appear to be SqlEntity not a generated type. And I don't want to specify columns one by one and return Tuple.

xperiandri:
Because providing type safe access to columns of entity is good for F# code but not interopability.

colinbull:
Sorry, missed your replies to this. 

So I guess you want to serialize the result directly from a call to the type provider? 

This isn't possible as the types that are provided by the type provider are erased at run-time. This essentially means that all of the metadata infrastructure that most serializers require is not available. 

However we have a work around by using the MapTo method on SqlEntity, see. [here for an example](https://github.com/fsprojects/SQLProvider/blob/master/tests/SqlProvider.Tests/scripts/OracleTests.fsx#L58), this simply works by matching properties on a name basis, but it is possible to override how this gets assigned.

Thorium:
If you do this:

```fsharp
let itms = query { ... } |> Seq.toArray //array of entities
let res = itms |> Array.map(fun i -> i.ColumnValues |> Map.ofSeq)
```

...and now you have quite nice array of maps that should be fairy simple to transfer to JSON (actually I think Newtonsoft can already serialize F#-map to JSON automatically).

PR #337, opened 2016-10-02
Map to records

s952163:
I was about to write a short mapping example but it seems there is already a mapper.fsx in docs/core so I just put it there with a .MapTo example and added the link to the main page. That said, ideally the type provider would return records of the row entity (or table) type so this mapping could be avoided. Is this something planned? Or technically not possible? Basically what came up in [convert-sqlprovider-typed-object-to-json-string](http://stackoverflow.com/questions/39611832/convert-sqlprovider-typed-object-to-json-string) and also [Issue 212](https://github.com/fsprojects/SQLProvider/issues/212). In that case one can skip on wiring up the initial types. For example the SQLDataConnection type provider in FSharp.Data does that. Although for SQLServer.


Thorium:
Thanks!

Documentation is always needed.

I commented also to other issue, I have never used MapTo, I do it like this:

```fsharp
let itms = query { ... } |> Seq.toArray //array of entities
let res = itms |> Array.map(fun i -> i.ColumnValues |> Map.ofSeq)
```

...and now you have quite nice array of maps that should be fairy simple to transfer to JSON (actually I think Newtonsoft can already serialize F#-map to JSON automatically).

s952163:
That's a nice approach. Let me add that too to the doc. I wonder though if it's gonna be an map of objects (will test later)?

s952163:
Thanks for the tip. I added the above example to mappers.fsx -> mappers.html. I think still returning a record directly from SQLEntity wold be nice. It's not only about serialization, but working with the data itself. Records are more natural. The map has its advantages as a column can be directly extracted.

pezipink:
It is not currently possible for type providers to return any special F# types, this is why we included MapTo, since at least it can do the mapping for you once you write the records.

s952163:
Understood. Thanks. Yes, `MapTo` will indeed get the job done. I wish I would've found it earlier. :-) One last question though, SqlDataConnection and also SqlCommandProvider do return a sequence of records. I believe those are also erasing type provider.

pezipink:
SqlDataConnection is is just a wrapper over SQLMetal.exe which is the code generation tool used by C# and Linq to SQL.  It is not really an erasing type provider, the types it gives you are full .NET types generated into an assembly thus mitigating the benefits of erased types.  These types are not F# record types.

SqlCommandProvider also does not return F# record types, it uses some trickery and DynamicObject to approximate something similar, see http://fsprojects.github.io/FSharp.Data.SqlClient/output.html for more information.

pezipink:
If you really REALLY want automatic record types, I wrote a [metaprogram ](https://github.com/pezipink/MixinProvider/blob/master/src/MixinProvider/TestMetaprograms/SQLRecords.fsx)for the [mixin type provider](https://github.com/pezipink/MixinProvider), that analyses your program and finds all usage of erased types from the SQLProvider. It then generates and injects record types and extension methods to convert to and from SQLEntity - I would not recommend you use this though it is more of a crazy experiment than anything else :)

s952163:
Thanks for the explanation. The mixin type provider looks fascinating, I will certainly explore it. In the meantime, `MapTo` it is!

PR #342, opened 2016-10-20
Update document of Owner filed in static parameter

JacobChang:
Owner field has different meanings, it's not Oracle Only any more


Thorium:
Thanks!

PR #345, opened 2016-10-21
Include VIEWs in table list returned for Oracle databases

### LINKED ISSUE #343: Oracle connections do not give access to VIEWs
When connecting to an Oracle database `view`s are not accessible, only real tables.

Unlike other databases (e.g. MSSQL, Postgres, and MySQL) Oracle makes a distinction between views and tables in it's information schema. [Querying `all_tables`](https://github.com/fsprojects/SQLProvider/blob/57346934cedc44ea4cbae5cdc11b007b669a8249/src/SQLProvider/Providers.Oracle.fs#L226) only returns real tables, information on accessible views must be retrieved separately from `all_views`. Alternatively the [`all_catalog` system table](https://docs.oracle.com/cd/B28359_01/server.111/b28320/statviews_1028.htm) can be queried, with results scoped to `table_type`s of `TABLE` and `VIEW`.

ah45:
Fixes #343

### DIFF
--- a/src/SQLProvider/Providers.Oracle.fs
+++ b/src/SQLProvider/Providers.Oracle.fs
@@ -222,9 +222,10 @@ module internal Oracle =
 
     let getTables tableNames conn = 
         let whereTableName = buildTableNameWhereFilter "table_name" tableNames
-        sprintf """select owner, table_name, tablespace_name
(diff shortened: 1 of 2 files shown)


Thorium:
Thanks!

PR #346, opened 2016-10-21
Provide Dockerfile to run a test Oracle DB and alter test script to run against it

ah45:
Testing changes to the Oracle provider is currently all but impossible outside of writing a script specific to whatever environment you have available at the time. The test script included in the repo is a good example of this: it hard codes the server name and even the resolution path of the Oracle DLL to the environment in which it was first written.

There seems to be a precedent for using Docker to provide a common testing environment, for MySQL, so I've created a Dockerfile that instantiates an Oracle DB and populates it with the test data (and functions, procedures, etc.)

Creating a container from this script takes a few minutes so I've also amended the test script so that it can be run repeatedly against the same instance without errors (some of the insertions it makes will trip the uniqueness constraints otherwise.)


Thorium:
Wow, cool, thanks!
Should we integrate this to our CI? Ping @colinbull

pezipink:
This is really awesome, thanks a ton!

colinbull:
Yep, definitely, that is what I was intending to do for the MySQL one. The only problem was I couldn't find a way of running the docker container in the CI build. Do Travis / AppVeyor support it now? But it is good never the less to have it local as well. We should do it for postgres also.

Issue #339, opened 2016-10-07
nicePascalName/niceCamelCase strips out first number after single letter


kellerd:
### Description

nicePascalName/niceCamelCase provides a strange name resolution to some of my tables/fields.  Our standard, archaic as it may be are names like 

ZC963_AUDIT_LOG
TC001_USERS
A21_SERVICE

_Just logging this here for tracking as it would have wide scoping breaking changes if corrected._
### Repro steps

Table name, stored procedure, column,ect with single letter (uppercase) followed by numbers will strip numbers of first digit
### Expected behavior

> nicePascalName "Z963_TABLESPACE" 
> val it : string = "Z963Tablespace"
> nicePascalName "A21_SERVICE"
> val it : string = "A21Service"
### Actual behavior

> nicePascalName "Z963_TABLESPACE" 
> val it : string = "Z63Tablespace"
> nicePascalName "A21_SERVICE"
> val it : string = "A1Service"
### Known workarounds

More than one character before the number

> nicePascalName "TC001_USERS"
> val it : string = "Tc001Users"
> nicePascalName "ZC963_AUDIT_LOG"
> val it : string = "Zc963AuditLog"

Lowercase first letter

> nicePascalName "z963_TABLESPACE"
> val it : string = "Z963Tablespace"
### Related information
- Operating system : Windows
- Branch : master
- .NET Runtime, CoreCLR or Mono Version : .NET Runtime

Thorium:
Fixed in version 1.0.39.

PR #355, opened 2016-11-02
Fix for #339: "AA21_AA" -> "Aa21Aa" instead of  "Aa1Aa"

Thorium:
Someone had copied code from the FSharp.Data project expecting it to work...

The pascalCase method cut some numbers from names.


pezipink:
we should make sure it mentions in the release note that this is potentially a breaking change.

pezipink:
I'm gonna blame @colinbull for this one :)

colinbull:
Guilty :)

Thorium:
Are you favour of using major version numbers (1.1 as opposed to 1.0.37) as a technical sign of breaking changes, or should those be used as business sign that the software should have some level of maturity?

PR #359, opened 2016-11-08
Use https for external js and css

Leonqn:
When viewing gh pages throught https, browser blocks downloading external scripts throught http.


Thorium:
Thanks!

PR #360, opened 2016-11-13
SQLite: Documentation about interop libraries.

o7g8:
Improved documentation about how to place SQLite interop libraries correctly on Windows and Linux platforms.

The documentation helps users to avoid following build error:

> The type provider 'FSharp.Data.Sql.SqlTypeProvider' reported an error: SQLite.Interop.dll, Path: <...>

(which is particularly misleading on Linux platform, where `libSQLite.Interop.so` is needed, but not `SQLite.Interop.dll`).


Thorium:
Thanks!

Issue #371, opened 2017-02-01
SQLite > ignored columns types


j-u-l-i-e-n:
### Description

SQLite provider doesn't work for types with a length constraint when there is a space between the type of the column and the length constraint, e.g. varchar (255)

### Repro steps

```
CREATE TABLE country (
    id            INTEGER       PRIMARY KEY AUTOINCREMENT
                                NOT NULL,
    name          VARCHAR (255) NOT NULL,
    area_id       INTEGER,
    local_currency_id   INTEGER,
    comments      TEXT,
    numeric_code  INTEGER (3)   UNIQUE,
    alpha2_code   VARCHAR (2)   UNIQUE,
    alpha3_code   VARCHAR (3)   UNIQUE,
    svg_data_url  TEXT
);
```
```

#r @"..\packages\SQLProvider.1.0.37\lib\FSharp.Data.SqlProvider.dll"

open FSharp.Data.Sql

let [<Literal>] resolutionPath = __SOURCE_DIRECTORY__ + @"\..\packages\System.Data.SQLite.Core.1.0.103\lib\net45" 
let [<Literal>] connectionString = "Data Source=" + __SOURCE_DIRECTORY__ + @"\northwindEF.db; Version=3; foreign keys=true"
// create a type alias with the connection string and database vendor settings
type sql = SqlDataProvider< 
              ConnectionString = connectionString,
              DatabaseVendor = Common.DatabaseProviderTypes.SQLITE,
              SQLiteLibrary = Common.SQLiteLibrary.AutoSelect,
              ResolutionPath = resolutionPath,
              IndividualsAmount = 1000,
              UseOptionTypes = true, 
              CaseSensitivityChange = Common.CaseSensitivityChange.ORIGINAL>
let db = sql.GetDataContext()

for x in db.Main.Country do
  printfn "%A" x. //columns are ignored depending on their type
```

I would exepct the column "Name" to appear in the list but it doesn't.

### Known workarounds

To make it work, it would require trimming the data type after line 267 in Providers.SQLite.fs, see below :

```
let dt = if dt.Contains("(") then dt.Substring(0,dt.IndexOf("(")) else dt
let dt = dt.Trim()
```

Julien

PR #372, opened 2017-02-01
added a call to trim the data type description from the table_info call

j-u-l-i-e-n:
see https://github.com/fsprojects/SQLProvider/issues/371


Thorium:
Thanks!

PR #377, opened 2017-02-14
added an update example

hermesxgjini:


docs/content/core/crud.fsx
```diff
@@ -121,6 +121,27 @@ mvps2
 
 ctx.SubmitUpdates()
 
+(** update a single row 
+*)
+
+type Employee2 = {
```
Thorium: You don't need this record to anything

docs/content/core/crud.fsx
```diff
@@ -121,6 +121,27 @@ mvps2
...
+    LastName:string
+}
+
+let updateEmployee (employee: Employee2) =
+    query {
+        for p in ctx.Public.Employee2 do
+        where (p.Id = employee.Id)
+    }
+    |> Seq.iter( fun e ->
+        e.FirstName <- post.FirstName
```
Thorium: What is post?

docs/content/core/crud.fsx
hermesxgjini: I added for the updateEmployee function because the Employee record above didn`t have an Id field

Thorium:
Thanks!

PR #379, opened 2017-02-26
Update index.fsx with more relevant Npgsql advice

damageboy:
The npgsql blurb is outdated, and the advice given in it is not longer true for the 3.x line

### DIFF
--- a/docs/content/index.fsx
+++ b/docs/content/index.fsx
@@ -36,7 +36,7 @@ All database vendors except SQL Server and MS Access will require 3rd party ADO.
 
 SQLite is based on the .NET drivers found [here](http://system.data.sqlite.org/index.html/doc/trunk/www/downloads.wiki). You will need the correct version for your specific architecture and setup.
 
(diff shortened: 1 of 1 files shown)


pezipink:
Thanks for this, I think a lot of people still use 2.x, is the new text applicable there as well?

damageboy:
2.x is completely on life support, I dare no think of the stuff that was
going on there. 3.x has been out for more than a year now, and is the only
thing being actively maintained. It's also the only version with net core
support. The text is definitely not relevant for 2.x

roji:
Yeah, Npgsql 2.x very officially unsupported (3.0 came out in August 2015, am the maintainer).

I admit I find it hard to understand how disabling pooling (`Pooling=false`) would help with being denied frequent access, even in 2.x. If anyone encounters something like this in 3.x I'd definitely be interested in hearing about it. At the very least, the comment should make it clear that the suggestion is only valid for 2.x.

damageboy:
I should also point out that you own [paket.lock](https://github.com/fsprojects/SQLProvider/blob/master/paket.lock#L38) is set to using 3.1.7 for which this advice is clearly wrong

pezipink:
fair enough

PR #382, opened 2017-03-10
Add GetDataContext overloads for passing in the TransactionOptions

anilmujagic:
Default transaction isolation level for `TransactionScope` is `Serializable`, which can easily cause transaction deadlocks. This provides option to explicitly set the `TransactionOptions` for `TransactionScope`, which is created on `SubmitChanges`.

**UPDATE:** I'm thinking if I've made a mistake by not keeping parameter order the same for added overloads and put `transactionOptions` as last one?

### DIFF
--- a/src/SQLProvider/Utils.fs
+++ b/src/SQLProvider/Utils.fs
@@ -55,17 +55,19 @@ module internal Utilities =
         | [] -> async { () }
 
 
-    let ensureTransaction() =
+    let ensureTransaction (transactionOptions : Transactions.TransactionOptions) =
         // Todo: Take TransactionScopeAsyncFlowOption into use when implemented in Mono.
         // Without it, transactions are not thread-safe over threads e.g. using async can be dangerous)
         // However, default option for TransactionScopeOption is Required, so you can create top level transaction
         // and this Mono-transaction will have its properties.
+        let transactionScopeOption = Transactions.TransactionScopeOption.Required
+
         let isMono = Type.GetType ("Mono.Runtime") <> null
         match isMono with
-        | true -> new Transactions.TransactionScope()
+        | true -> new Transactions.TransactionScope(transactionScopeOption, transactionOptions)
         | false ->
             // Note1: On Mono, 4.6.1 or newer is requred for compiling TransactionScopeAsyncFlowOption.
-            new Transactions.TransactionScope(System.Transactions.TransactionScopeAsyncFlowOption.Enabled)
+            new Transactions.TransactionScope(transactionScopeOption, transactionOptions, Transactions.TransactionScopeAsyncFlowOption.Enabled)
 
     let parseAggregates fieldNotation fieldNotationAlias query =
         let rec parseAggregates' fieldNotation fieldNotationAlias query (selectColumns:string list) =
--- a/src/SQLProvider/Providers.MSAccess.fs
+++ b/src/SQLProvider/Providers.MSAccess.fs
@@ -468,7 +468,7 @@ type internal MSAccessProvider() =
             let sql = sb.ToString()
             (sql,parameters)
 
-        member this.ProcessUpdates(con, entities) =
+        member this.ProcessUpdates(con, entities, transactionOptions) =
             let sb = Text.StringBuilder()
 
             entities.Keys |> Seq.iter (fun e -> printfn "entity - %A" e.ColumnValues)
@@ -520,7 +520,7 @@ type internal MSAccessProvider() =
                 ()
                 //con.Close()
 
-        member this.ProcessUpdatesAsync(con, entities) =
+        member this.ProcessUpdatesAsync(con, entities, transactionOptions) =
             let sb = Text.StringBuilder()
 
             entities.Keys |> Seq.iter (fun e -> printfn "entity - %A" e.ColumnValues)
--- a/src/SQLProvider/Providers.MsSqlServer.fs
+++ b/src/SQLProvider/Providers.MsSqlServer.fs
@@ -742,7 +742,7 @@ type internal MSSqlServerProvider(tableNames:string) =
             let sql = sb.ToString()
             (sql,parameters)
 
-        member this.ProcessUpdates(con, entities) =
+        member this.ProcessUpdates(con, entities, transactionOptions) =
             let sb = Text.StringBuilder()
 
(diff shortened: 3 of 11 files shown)


src/SQLProvider/Utils.fs
```diff
@@ -55,17 +55,19 @@ module internal Utilities =
...
         // Todo: Take TransactionScopeAsyncFlowOption into use when implemented in Mono.
         // Without it, transactions are not thread-safe over threads e.g. using async can be dangerous)
         // However, default option for TransactionScopeOption is Required, so you can create top level transaction
         // and this Mono-transaction will have its properties.
+        let transactionScopeOption = Transactions.TransactionScopeOption.Required
+
         let isMono = Type.GetType ("Mono.Runtime") <> null
         match isMono with
-        | true -> new Transactions.TransactionScope()
+        | true -> new Transactions.TransactionScope(transactionScopeOption, transactionOptions)
```
pezipink: I know you didn't do this, but is this code actually correct? it looks around the wrong way to me, surely the true branch should have the reflection stuff in it?  @Thorium ?

pezipink: Ah, my bad, I think I was looking at an old version of the file for some reason!

anilmujagic: Are you talking about the code which was there before @Thorium 's commit, which used the reflection? I understood it is there because if we're not in Mono (isMono = false) then we want to use the feature available in .NET, but we have to use reflection because that enum doesn't even exist in Mono and it wouldn't compile.

pezipink: Oh right, I see. Yes, that is what I was talking about.  Clearly I did not read it properly :)

anilmujagic: Any thoughts on my update? Shall I reorder params?

pezipink:
Looks ok to me in general, a couple of thoughts though:

* Do we really need the extra static parameter? There's already loads of them and you can set this at runtime anyway.

* We could make the transaction options get/settable on the data context so that the isolation level can be changed for different calls on the same context 

WDYT?

anilmujagic:
I thought to do it that way (it'd be cleaner and simpler to implement after all), but went for extending GetDataContext method for two reasons:
1. Thread safety
2. *Functional way*

But I guess if I use `lock` for 1st and close one eye on 2nd, it'll be fine ;-). OK, I'm going to change the implementation and update the pull request.

Thorium:
What it comes to me removing the reflection, mono has the AsyncFlow-type in 4.6.1-... and 4.8-... so we don’t need reflection anymore, that is great. But why did our travis CI fail, not having the AsyncFlow-type??

I also think the isolation level should be non-static. But why the default “commit to existing transaction” wasn’t enough? Then when the user uses transaction isolation he/she wants, we commit to that.

anilmujagic:
@Thorium It failed on Mono installation step: `The command "sudo apt-get install -qq mono-complete mono-vbnc fsharp nuget referenceassemblies-pcl" failed and exited with 100 during .`

anilmujagic:
@Thorium On transaction isolation level: I would need to have TransactionScope all over the place in the code, where I'm dealing with context. I think it's better to solve that on the context level, and then all I need is one function:
```fsharp
    let getDataContext() =
        let options = new Transactions.TransactionOptions(IsolationLevel = Transactions.IsolationLevel.ReadCommitted)
        Database.GetDataContext(options)
```

pezipink:
Yes, good point about the thread safety.  As for _funcitonal_ way, don't worry about that, the SQLProvider provides a non functional experience anyway :)

Thorium:
Hi,

GetDataContext() is not using transactions nor isolation levels anyway, so that would be tricking the developer, right? The only command that is using them is Update().

Now, let's look this:

```fsharp
let context = Database.GetDataContext cstr
use tr = 
    new Transactions.TransactionScope(
        Transactions.TransactionScopeOption.Required, 
        new Transactions.TransactionOptions(
            IsolationLevel = Transactions.IsolationLevel.ReadCommitted), 
        Transactions.TransactionScopeAsyncFlowOption.Enabled)
entity1.SetColum("a", 1)
context.SubmitUpdates()
entity2.SetColum("b", 2)
context.SubmitUpdates()
tr.Complete()
```
the transaction level inside context is Required so entity1 and entity2 will commit to the tr transaction with the isolation level ReadCommitted.

You wouldn't like to make just a single `Database.GetDataContext()` anyway so you still have to call `Database.GetDataContext(options)` everywhere. Or how do you deal with this: if you have multiple users and one of them feed invalid data to context so that some database-operation will fail, you would have to call `ClearUpdates()` which would clear everyone elses data also.

anilmujagic:
@Thorium Of course I don't have one instance of database context, but I use one single "factory" function to create it when I need a new one. That way the construction of the new context is isolated into one single place. What you're suggesting would lead me to polluting every single data access function with `TransactionScope`.

Thorium:
I agree that "factory methods" are polluting.

```
let doStuff() =
   writeWithDbContext <| fun (dbContext:DataContext) ->
      //here the normal code that uses dbContext
```
where `writeWithDbContext` is a "factory method", it creates the database-context and wraps code to a transaction, and executes transaction `Complete()` if everything went fine.

anilmujagic:
Yes, that would be an option. However, I still think providing the option to set it on the Context level is valuable, and simpler from the developer perspective. It doesn't require you to change the code to apply different isolation level in the case you need it later in the project life.

I had this before in the C# project, and only when we started having high load, then deadlocks started to occur. Being able to set it on one single central place, where we create EF context instance, made the fix very simple. If we have had to wrap hundreds of DB calls in wrapper function, that would make the fix take days to implement.

Thorium:
Oh no... Can we do anything without adding the System.Transactions.dll to everywhere...?

image

anilmujagic:
Oh, it's merged already. @pezipink shall I still change it to property (what we discussed above)? If no, then shall I reorder the parameters as mentioned in my initial comment *update* part?

Thorium:
I didn't create a new package yet, so we can do more merges...

anilmujagic:
@Thorium Ugh, you're right. How about having transaction isolation level as a union type and have that mapped internally?

Thorium:
If it helps. Currently the SQLProvider.Tests.sln is not having reference to System.Transactions, so that's easy to test.

PR #383, opened 2017-03-11
Avoid forcing dependency on System.Transactions

anilmujagic:
I've created corresponding types for TransactionOptions and IsolationLevel, which are internally mapped to System.Transactions types.

Sample usage:

```fsharp
open FSharp.Data.Sql
open FSharp.Data.Sql.Transactions

type Database = SqlDataProvider<ConnectionStringName = "DB">

let options = { TransactionOptions.Default with IsolationLevel = IsolationLevel.ReadCommitted }
let context = Database.GetDataContext(options)
```

@pezipink The more I think about having these options as a mutable property on the context, the more I don't like it :(. Using the context from multiple threads (which should be avoided anyway) and setting options from those threads, can lead to really unpredictable results. By having it passed to the GetDataContext we're forcing the dev to create separate contexts if (s)he wants to have transactions hadled differently.

### DIFF
--- a/src/SQLProvider/Providers.MsSqlServer.fs
+++ b/src/SQLProvider/Providers.MsSqlServer.fs
@@ -6,6 +6,7 @@ open System.Collections.Generic
 open System.Data
 open System.Data.SqlClient
 open FSharp.Data.Sql
+open FSharp.Data.Sql.Transactions
 open FSharp.Data.Sql.Schema
 open FSharp.Data.Sql.Common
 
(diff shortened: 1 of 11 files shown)


Thorium:
Thanks!

Thorium:
I guess our other [static parameters](http://fsprojects.github.io/SQLProvider/core/parameters.html) are also affecting design time, so we can't move them to dbcontext options?

anilmujagic:
My comment was about pezipink's suggestion to have the TransactionOptions as a property of DataContext, as it affects the runtime only.

Thorium:
No, the current is fine. :-)

Issue #394, opened 2017-04-20
Oracle DELETE causes ORA-00932 for PK of a type NUMBER(10,0)


object:
### Description

I have an Oracle database with tables that have surrogate primary keys of a type NUMBER(10,0).
When I try to delete a row in such table, Oracle provider throws an exception:

**ORA-00932: inconsistent datatypes: expected NUMBER, got INTERVAL YEAR TO MONTH**

### Repro steps

1. Define a table with a primary key of a type NUMBER(10, 0)

2. Retrieve a row in that table using SQLProvider

3. Call row.Delete()

4. Call db.SubmitUpdates()

### Expected behavior

The row should be deleted.

### Actual behavior

An Oracle ORA-00932 exception as shown above 

### Known workarounds

I will submit a PR with a fix shortly.

object:
PR was accepted, closing the issue.

PR #395, opened 2017-04-20
Fix for #394 (ORA-00932 when deleting rows in Oracle).

object:
I didn't fully understand the failing code that generated Oracle command (not sure why it had to call findClrType but obviously the type wasn't correct), but createUpdateCommand function contained perfecly working code for Oracle command generation, so I copied the part of it that was relevant for deletion and then everything worked just fine.

### DIFF
--- a/src/SQLProvider/Providers.Oracle.fs
+++ b/src/SQLProvider/Providers.Oracle.fs
@@ -522,16 +522,13 @@ type internal OracleProvider(resolutionPath, owner, referencedAssemblies, tableN
         | [] -> ()
         | ks -> 
             ~~(sprintf "DELETE FROM %s WHERE " entity.Table.FullName)
-            ~~(String.Join(" AND ", ks |> List.mapi(fun i k -> (sprintf "%s = :id%i" k i))))
+            ~~(String.Join(" AND ", ks |> List.mapi(fun i k -> (sprintf "\"%s\" = :pk%i" k i))))
 
         let cmd = provider.CreateCommand(con, sb.ToString())
-        cmd.CommandType <- CommandType.Text
         pkValues |> List.iteri(fun i pkValue ->
-            let pkType = pkValue.GetType().ToString();
-            match Oracle.findClrType pkType with
-            | Some(m) ->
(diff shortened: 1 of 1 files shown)


Thorium:
Thanks!

Thorium:
Hmm. Does `.ToString()` generate correct values on basic cases, if e.g. the ID is a Guid (RAW(16) type)?

object:
Not sure I understood the question.ToString on what object?

Thorium:
Sorry, I meant that command parameter type, is now String, when it was dynamically the type that corresponded the type of CLR. But my question is: 

> Does this new code work on tables that have primary key of type RAW(16)?

object:
I will check this tomorrow at work where I have Oracle installed.

object:
Checked now, everything works fine. Here's what I did.

1. Defined a new table:

CREATE TABLE "INTEGRATION"."TEST_GUIDS" ( "GUID_ID" RAW(16) NOT NULL ENABLE, "SOME_TEXT" NVARCHAR2(50) NOT NULL ENABLE, PRIMARY KEY ("GUID_ID"));

2. Ran the following code:

let row = dbContext.Integration.TestGuids.Create("ABC")
row.GuidId <- Guid.NewGuid().ToByteArray()
dbContext.SubmitUpdates()
row.SomeText <- "XYZ"
dbContext.SubmitUpdates()
row.Delete()
dbContext.SubmitUpdates()

3. Checked that the row was created, then updated, then deleted. Here are commands executed:

> 
Executing SQL: INSERT INTO INTEGRATION.TEST_GUIDS (SOME_TEXT,GUID_ID) VALUES (:param0,:param1)
val it : Unit = ()
> 
Executing SQL: UPDATE INTEGRATION.TEST_GUIDS SET (SOME_TEXT) = (SELECT :param0 FROM DUAL) WHERE "GUID_ID" = :pk0
val it : Unit = ()
> 
Executing SQL: DELETE FROM INTEGRATION.TEST_GUIDS WHERE "GUID_ID" = :pk0
val it : Unit = ()
>

Issue #401, opened 2017-04-28
Easier way to delete multiple records


Thorium:
Instead of writing a select and then delete-one-by-one, I want this:

```fsharp
dbContext.myTable.Delete(fun x -> x.Thing = 1)
```
to generate this

```sql
DELETE FROM [myTable] WHERE [myTable].[Thing] = 1
```

We currently have dbContext.myTable.Create() so why not also .Delete()?

Thorium:
...or maybe more LINQ-style would fit for us, something like:
```fsharp
query {
  for i in dbContext.myTable do
  where (fun x -> x.Thing = "Hello Ross")
} |> DeleteAll
```

Deletion is not a good practice with databases (as setting a Deleted-flag is considered better), so it shouldn't be too easy.

Thorium:
DELETE queries cannot have as-alias like `delete from customer as cust where cust.Name = ''`
and we generate as-alias always. So the design from the first message could be better.

PR #404, opened 2017-05-05
Possibility to use simple but dangerous DELETE-queries. 

Thorium:
You have done backups of your database, right?

This is for feature #401

```fsharp
query {
    for c in ctx.Dbo.Employees do
    where (...)
} |> Seq.``delete all items from single table``  |> Async.RunSynchronously
```
```sql
DELETE FROM [dbo].[EMPLOYEES] WHERE (...)
```

Be warned, you will shoot your own foot with this cannon.

Happy deleting!


Thorium:
I'm not totally happy to put the method under Seq-module, but I can't find any better place either.

PR #409, opened 2017-05-09
Add support for casts to IComparable in lambdas of sortBy, sortByDescending, thenBy and thenByDescending query operations.

### LINKED ISSUE #407: Allow casts to IComparable in Lambdas passed to sortBy or sortByDescending.
Currently, the following query fails with an [unrecognized method call exception](https://github.com/fsprojects/SQLProvider/blob/3ebcceb2a4a15915c30178e762ae9ec451a8fc73/src/SQLProvider/SqlRuntime.Linq.fs#L753),
because the cast to `IComparable` inserts a `Convert` node into the expression which is not matched:

```fsharp
query {
    for item in db.Main.Items do
    sortBy ((fun (i : Db.dataContext.``main.ItemsEntity``) -> m.Id :> IComparable) item)
    select item
}
```

I propose to change [this line handling `sortBy` and `sortByDescending`](https://github.com/fsprojects/SQLProvider/blob/3ebcceb2a4a15915c30178e762ae9ec451a8fc73/src/SQLProvider/SqlRuntime.Linq.fs#L480) from

```fsharp
| MethodCall(None, (MethodWithName "OrderBy" | MethodWithName "OrderByDescending" as meth), [SourceWithQueryData source; OptionalQuote (Lambda([ParamName param], SqlColumnGet(entity,key,_))) ]) ->
```

this to

```fsharp
| MethodCall(None, (MethodWithName "OrderBy" | MethodWithName "OrderByDescending" as meth), [SourceWithQueryData source; OptionalQuote (Lambda([ParamName param], OptionalConvertOrTypeAs (SqlColumnGet(entity,key,_)))) ]) ->
```

this. The only change is the insertion of the `OptionalConvertOrTypeAs` recognizer before the `SqlColumnGet` recognizer, which allows the query above to pass.
The same change should also be made to [the line handling `thenBy` and `thenByDescending`](https://github.com/fsprojects/SQLProvider/blob/3ebcceb2a4a15915c30178e762ae9ec451a8fc73/src/SQLProvider/SqlRuntime.Linq.fs#L495):

```fsharp
| MethodCall(None, (MethodWithName "ThenBy" | MethodWithName "ThenByDescending" as meth), [SourceWithQueryData source; OptionalQuote (Lambda([ParamName param], OptionalConvertOrTypeAs (SqlColumnGet(entity,key,_)))) ]) ->
```

**I should mention that the `Convert` node is only inserted for value types that have to be boxed!
Casting a `string` to `IComparable` does not affect the resulting expression.**

## Reason why this would be helpful:

Currently, it is not possible to splice in a lambda for `sortBy` or `sortByDescending` into a query:

```fsharp
// The cast to IComparable is necessary for type checking
// (Expr<Db.dataContext.``main.ItemsEntity`` -> IComparable).
let sortBy =
    match sortArg with
    | "name" -> <@ fun (x : Db.dataContext.``main.ItemsEntity``) -> m.Name :> IComparable @>
    | "id" -> <@ fun (x : Db.dataContext.``main.ItemsEntity``) -> m.Id :> IComparable @>
    | _ -> failwith "invalid arg to sort by"

let quotedQuery = <@ query {
    for item in db.Main.Items do
    sortBy ((%sortBy) item)
    select item
} @>
```

Would this be an acceptable enhancement and should I open a pull request for it?

nikonthethird:
Fixes #407.

This change ignores optional `Convert` expression nodes inside lambdas passed to `sortBy`, `sortByDescending`, `thenBy` or `thenByDescending` operations:

```fsharp
query {
    for item in db.Main.Items do
    sortBy ((fun (i : Db.dataContext.``main.ItemsEntity``) -> m.Id :> IComparable) item)
    select item
}
```

This query is now convertible to SQL.

I also added two unit tests (one for `sortBy` and one for `thenByDescending`) to check if the cast to `IComparable` is allowed.

### DIFF
--- a/src/SQLProvider/SqlRuntime.Linq.fs
+++ b/src/SQLProvider/SqlRuntime.Linq.fs
@@ -477,7 +477,7 @@ module internal QueryImplementation =
                         let ty = typedefof<SqlQueryable<_>>.MakeGenericType(meth.GetGenericArguments().[0])
                         ty.GetConstructors().[0].Invoke [| source.DataContext ; source.Provider; Take(amount,source.SqlExpression) ; source.TupleIndex; |] :?> IQueryable<_>
 
-                    | MethodCall(None, (MethodWithName "OrderBy" | MethodWithName "OrderByDescending" as meth), [SourceWithQueryData source; OptionalQuote (Lambda([ParamName param], SqlColumnGet(entity,key,_))) ]) ->
+                    | MethodCall(None, (MethodWithName "OrderBy" | MethodWithName "OrderByDescending" as meth), [SourceWithQueryData source; OptionalQuote (Lambda([ParamName param], OptionalConvertOrTypeAs (SqlColumnGet(entity,key,_)))) ]) ->
                         let alias =
                              match entity with
                              | "" when source.SqlExpression.HasAutoTupled() -> param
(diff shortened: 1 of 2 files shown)


nikonthethird:
That was quick, thank you!

PR #410, opened 2017-05-09
Use trim instead of RTRIM(LTRIM())

colinbull:


Thorium:
Why Oracle docs say this: https://docs.oracle.com/database/121/ODPNT/canonical_map.htm

colinbull:
While that is true `RTRIM(LTRIM())` should equal `TRIM()` I have experienced some interesting differences in behaviors under certain NLS_CHARACTERSETS with oracle, `TRIM()` seems to be better behaved, in my experience. I'm not sure why as I guess it calls RTRIM and LTRIM anyway, I guess just another oracle oddity, that I'll have to chalk up to experience.

Thorium:
Ok, maybe Trim does not exists in some very old Oracles...

PR #411, opened 2017-05-09
Firebird 2.x provider

gibranrosa:
* Modifications to SqlProvider common modules:

SqlRuntime.Common : 
	lines 192 and 210-213, unquoted alias prefix (toLower)
	line 25: FIREBIRD provider constant
SqlRuntime.DataContext :
	line 25: FIREBIRD provider creation

Utils :
	lines 311-357: 
		IDataReader wrapper that keeps the Command object in scope,
		because the Firebird reader gets closed if the Command gets out of scope
		* the Dispose method for that wrapper calls the Dispose for the Command
	line 406-407:
		change the 'use' to 'let'
		wrap the IDataReader in a IDataReaderWithCommand, as explained above.

* Added FirebirdTests.fsx to Tests Project


Thorium:
Very nice!
I'll do a little code review, but in general looks very great!

src/SQLProvider/SqlProvider.fsproj
```diff
@@ -107,7 +108,7 @@
   </Target>
   -->
   <Choose>
-    <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And ($(TargetFrameworkVersion) == 'v4.0' Or $(TargetFrameworkVersion) == 'v4.5' Or $(TargetFrameworkVersion) == 'v4.5.1' Or $(TargetFrameworkVersion) == 'v4.5.2' Or $(TargetFrameworkVersion) == 'v4.5.3' Or $(TargetFrameworkVersion) == 'v4.6' Or $(TargetFrameworkVersion) == 'v4.6.1' Or $(TargetFrameworkVersion) == 'v4.6.2' Or $(TargetFrameworkVersion) == 'v4.6.3' Or $(TargetFrameworkVersion) == 'v4.7')">
+    <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And ($(TargetFrameworkVersion) == 'v4.0' Or $(TargetFrameworkVersion) == 'v4.5' Or $(TargetFrameworkVersion) == 'v4.5.1' Or $(TargetFrameworkVersion) == 'v4.5.2' Or $(TargetFrameworkVersion) == 'v4.5.3' Or $(TargetFrameworkVersion) == 'v4.6' Or $(TargetFrameworkVersion) == 'v4.6.1' Or $(TargetFrameworkVersion) == 'v4.6.2' Or $(TargetFrameworkVersion) == 'v4.6.3')">
```
Thorium: This is just because the Paket version has been older...

src/SQLProvider/Utils.fs
```diff
@@ -384,8 +431,8 @@ module Sql =
         }
 
     let executeSql createCommand sql (con:IDbConnection) =        
-        use com : IDbCommand = createCommand sql con   
-        com.ExecuteReader()    
+        let com : IDbCommand = createCommand sql con
```
Thorium: What is the point of new DataReaderWithCommand-type?
How much longer will the command live in the memory when instead of disposing it, it is captured to this new type?

src/SQLProvider/Providers.Firebird.fs
```diff
@@ -0,0 +1,962 @@
...
+            match String.IsNullOrEmpty(al) with
+            | true -> sprintf "%s"
+            | false -> sprintf "%s.%s" al
+        match c with
+        // Custom database spesific overrides for canonical functions:
+        | SqlColumnType.CanonicalOperation(cf,col) ->
+            let column = fieldNotation al col
+            match cf with
+            // String functions
+            | Replace(SqlStr(searchItm),SqlStrCol(al2, col2)) -> sprintf "REPLACE(%s,'%s',%s)" column searchItm (fieldNotation al2 col2)
```
Thorium: These commands should correspond the database specific commands, in this case listed here: https://firebirdsql.org/refdocs/langrefupd21-intfunc.html

src/SQLProvider/Providers.Firebird.fs
```diff
@@ -0,0 +1,962 @@
...
+                        | Deleted | Unchanged -> failwith "Unchanged entity encountered in update list - this should not be possible!"
+
+                    do! Utilities.executeOneByOne handleEntity (entities.Keys|>Seq.toList)
+
+                    if scope<>null then scope.Complete()
+
+                finally
+                    con.Close()
+            }
+        (* member this.ProcessUpdatesAsync(con, entities) =
```
Thorium: What is this commented section here?

gibranrosa: My mistake, just thrash 🤓

src/SQLProvider/Utils.fs
Thorium: Utils.fs class is used also by other providers. But `executeSql` is not executed by any interfaces or any common code. So I think you can add your version of the function to be spesifically in the Providers.Firebird.fs and do the disposable-handling there, leaving the common code to the current version. E.g. there is custom `executeSql` in Providers.MsSqlServer.fs (I don't know why).

And in the Providers.firebird.fs, maybe you don't need the whole type, if you just create a holder to keep reference of the cmd as long as you want, e.g., something like:
```fsharp
//either mutable or not, I don't know the case exactly.
let mutable myCmd = Unchecked.defaultof<IDisposable>
//then when needed:
myCmd <- (com : IDbCommand)

//later:
if myCmd <> Unchecked.defaultof<IDisposable> then
      myCmd.Dispose()
```

gibranrosa: Utils.fs restored!

src/SQLProvider/Providers.Firebird.fs
gibranrosa: All done! I think now it is correct 👌

src/scripts/Script1.fsx
```diff
@@ -0,0 +1,45 @@
...
+let [<Literal>] resolutionPath = "D:/Gibran/Projetos/GitRep/SQLProvider/packages/scripts/FirebirdSql.Data.FirebirdClient/lib/net452" //"D:/Gibran/Projetos/FirebirdNetProvider141216/src/FirebirdSql.Data.FirebirdClient/bin/Debug/NET40" 
+let [<Literal>] connectionString = @"character set=NONE;data source=localhost;port=3051;initial catalog=d:\Tisul\Gestao\Dados\ROMENA.FDB;user id=SYSDBA;password=masterkey;dialect=1"
+// create a type alias with the connection string and database vendor settings
+
+              
+
+
+
+
+type sql = SqlDataProvider<
```
Thorium: Currently our test cases are at `/SQLProvider/tests/SqlProvider.Tests/scripts` by the name of the provider and file is included in SqlProvider.Tests.fsproj. Needed sample database scripts are at `src/DatabaseScripts`

pezipink:
amazing PR out of nowhere, thanks a ton!  @Thorium we should bump to 1.1 for this when you are happy !  💯

Thorium:
@gibranrosa, thanks! Merged and released.

gibranrosa:
@pezipink , thanks, in fact it was kind of easy using the Mysql provider as the starting point. 

@Thorium Thanks for the review, I have yet to standardize the tests, and the Procedures with parameters wasn't working yet too, I forgot to test this. Do I make another PR when I am ready?

Thorium:
Yes, you can make another PR/PRs. I haven't used Firebird so I didn't do any testing, but usually there are some minor things to fix with new providers, like binary-column-support, individuals-queries, procedure output parameters, identity-insert-returning, etc, providers are complex things so I don't expect it to be perfect at the first try, the release-cycle is easy and fast.

pezipink:
@gibranrosa cool, yes, I designed it from the start so it would be as easy as possible to implement a new provider based on an existing one.  A perhaps interesting historical factoid, I wrote SQL server then basically copied it to create sqlite, mysql and postgres.  I fiddled around with them, tweaked the generated SQL, and got them to a baseline of working - then over  the years they have been extensively re-worked by people who understand those databases, especially around tricky things like type mapping (for postgres) and sprocs.  So, as @Thorium says, the Firebird side will likely go through a similar cycle now as it starts to be used.

PR #414, opened 2017-05-13
Firebird: Stored Procedures, tests and general fixes

gibranrosa:
Stored procedures now working, tests standardized, and more fixes to Firebird provider.


Thorium:
Looking good! Thanks.

PR #415, opened 2017-05-15
Mysql: changed the SQL queries which gets the Relationships and Columns

gibranrosa:
To make them more efficient, querying less tables from the Information_schema. We have a server with many databases and tables, and the old queries were taking too long. Besides the relationships query was not working with Mysql < 5.5, now it works with 5.0.6 until 5.6 at least. The test script for Mysql passed totally for Mysql 5.6.


Thorium:
Thanks!

Issue #420, opened 2017-05-23
CRUD operations won't serialize the data parameters


Thorium:
If I want to log the SQL-events I can do it like this:

```fsharp
FSharp.Data.Sql.Common.QueryEvents.SqlQueryEvent |> Event.add (printfn "Executing SQL: %s")
let filteredQuery =
    query { for customer in ctx.Main.Customers do
            where (customer.ContactName = "John Smith")
            select customer } |> Seq.toArray
```
and the parameters are in the end:

```sql
Executing SQL: SELECT [customer].[Address] as 'Address',[customer].[City] as 'City',[customer].[CompanyName] as 'CompanyName',[customer].[ContactName] as 'ContactName',[customer].[ContactTitle] as 'ContactTitle',[customer].[Country] as 'Country',[customer].[CustomerID] as 'CustomerID',[customer].[Fax] as 'Fax',[customer].[Phone] as 'Phone',[customer].[PostalCode] as 'PostalCode',[customer].[Region] as 'Region' FROM main.Customers as [customer] WHERE (([customer].[ContactName] = @param1)) - params @param1 - "John Smith"; 
```

But CRUD-operations are missing the parameters.

Now that I think of it, the event shouldn't be a string but a type that would carry both the query and the parameters. And for backward compatibility, the type should override .ToString() to display the query as now.

pezipink:
sounds good to me

PR #421, opened 2017-05-23
Executing SQL -event refactored #420

Thorium:
Executing SQL -event refactored #420
Also I noticed I didn't push previous release notes for 1.1.2 so here they are.


pezipink:
Nice 👍

Issue #437, opened 2017-07-14
MariaDb (Mysql) int column is mapped to unit32 data type


xuanduc987:
### Description

MariaDb int column is mapped to unit32 data type, causing overflow exception in runtime

![image](https://user-images.githubusercontent.com/1186411/28205268-40ea42ca-68bc-11e7-95a7-141f1cc4a24a.png)

### Repro steps

1. Create DB with int column in MariaDb

2. Create a record with negative value

3. Access that record via SQLProvider Mysql

### Expected behavior

Get value normaly

### Actual behavior

```
System.OverflowException: Value was either too large or too small for a UInt32.
   at System.Convert.ToUInt32(Int32 value)
   at System.Int32.System.IConvertible.ToUInt32(IFormatProvider provider)
   at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
   at System.Convert.ChangeType(Object value, Type conversionType)
   at FSharp.Data.Sql.Common.SqlEntity.GetColumn[T](String key)
   at <StartupCode$FSI_0004>.$FSI_0004.main@()
Stopped due to error
```

### Known workarounds

None yet

### Related information 

* Used database: MariaDb 10.2.6
* Operating system Windows 10
* Branch
* .NET Runtime, CoreCLR or Mono Version: 4.5.2
* Performance information, links to performance testing scripts

Thorium:
I totally agree, MySql `INT` should be mapped to .NET `int` and MySQL `INT UNSIGNED` should be mapped to .NET `UInt32`. Just to note, change to this is a major breaking change for existing users.

I think the current mapping comes from MySQL.Data library reference directly and is not overridden in our code...?

V1nzi:
I have a similar issue where all my primary keys (which are all int) are mapped to uint32. This is not a problem until I try to map the database entities (including the primary key) to a domain model using the .MapTo-method as described in [this part](http://fsprojects.github.io/SQLProvider/core/mappers.html) of the documentation.
It seems like the column is actually mapped to int32 at runtime, but uint32 at compile time, because I get a ConvertException saying that int32 cannot be converted to uint32 when trying to run my project.

One workaround would probably be to just create a custom map function and convert the datatypes as needed.

### Related information

- database: MariaDb 10.2.7
- OS: Arch Linux
- .NET runtime: Mono 5.0.0
- target framework: .NET 4.6.1
- using Monodevelop as IDE

Thorium:
[...]

PR #444, opened 2017-09-07
MYSQL Breaking change: Integers are not mapped to UInt32 but int.

Thorium:
INT USIGNED are mapped to UInt32.


Thorium:
This will fix issue #437

PR #448, opened 2017-09-14
Separate FSharp.Data.SqlProvider.dll and FSharp.Data.SqlProvider.DesignTime.dll

Thorium:
...this way we will get smaller runtime dll.

The difference:
Old FSharp.Data.SqlProvider.dll 1.84 MB
New FSharp.Data.SqlProvider.dll 1.47 MB

This PR include:
- Separate FSharp.Data.SqlProvider.dll and FSharp.Data.SqlProvider.DesignTime.dll
- commandTimeout parameter (PR #447)
- The new way of using ProvidedTypesContext.fs from FSharp.TypeProviders.StarterPack
- Build of netstandard2.0 version of FSharp.Data.SqlProvider.dll but not FSharp.Data.SqlProvider.DesignTime.dll and not packaging it.


Thorium:
I was thinking that separating the design time dll and runtime dll could also help a) debugging when Visual Studio locks the file, b) .NET standard scenario where the desing-time dll maybe could be .NET45 while the runtime could be core. But after a while of testing I'm not sure anymore that will this help at all. Because it's kind of a circular reference: the design time dll will need the runtime, and the runtime has to have attribute to instruct VS, but when building the runtime, the desing-time dll is not present yet.

Thorium:
Let's not go for this yet, I'll do some smaller initial changes first...

Issue #378, opened 2017-02-19
Create longer version does not contains the primary key


diegobfernandez:
### Description

The ``Create()`` method does not contains the table primary key.

### Repro steps

1. Create table
```
CREATE TABLE unidadeDeMedida (
  codigo INT PRIMARY KEY,
  descricao NVARCHAR(250) NOT NULL
)
```
2. Define type
```
type Db =
    SqlDataProvider<Common.DatabaseProviderTypes.MYSQL,
                    DesignConnectionString,
                    ResolutionPath = ResolutionPath,
                    UseOptionTypes = true,
                    CaseSensitivityChange = CaseSensitivityChange.ORIGINAL>
``` 
3. Try to access ``Create(codigo, descricao)``
```
let InserirUnidadesDeMedida connectionString unidadesDeMedida =
  let ctx = Db.GetDataContext connectionString
  unidadesDeMedida
  |> Seq.map (fun x ->
    ctx.Mdic.UnidadeDeMedida.``Create(codigo, descricao)``(x.Codigo, x.Descricao))
  |> ignore

  ctx.SubmitUpdates ()
```

### Expected behavior

The Create(codigo, descricao) should exist.

### Actual behavior

Compile error
```
FS0039: The field, constructor or member 'Create(codigo, descricao)' is not defined
```

### Known workarounds

Do not use Create(...) method

### Related information 

* Used database: MariaDB 10.1.21
* Operating system: Fedora 25
* Branch: The one from Nuget (1.0.42)
* .NET Runtime, CoreCLR or Mono Version: Mono 4.6.2.16

Thorium:
Another workaround is to generate ids in database:
```sql
CREATE TABLE unidadeDeMedida (
  `codigo` INT NOT NULL AUTO_INCREMENT COMMENT '',
  `descricao` NVARCHAR(250) NOT NULL,
  PRIMARY KEY (`codigo`)  COMMENT ''
)
```

diegobfernandez:
@Thorium I tried your script and it still does not work.
I forgot to mention there is only a method `Create(descricao)` in both cases.

Thorium:
Yes, at least in the later case there shouldn't be codigo at all, because it's database generated, so you don't insert primary key manually but database will generate it when the item is inserted. You are right that in the first case you should be able to insert a primary key, but I don't know if we know if primary keys are generated or not. Anyways you can use the method missing the primary key and then try to set it later, like

```fsharp
[...]

PR #450, opened 2017-09-19
Support arrays in PostgreSQL

piaste:
Enables SQLProvider to recognize array columns in PostgreSQL.

PostgreSQL allows you to specify both array rank and size, but enforces neither. This PR recognises and enforces array rank, but neither displays nor enforces array size.

If an array column is used as a primary key, it can also be used to browse through `.Individuals`. However, this currently does not work for multidimensional arrays. Debugging, the issue seems to be that the array goes into [the `GetterCode` quotation](https://github.com/piaste/SQLProvider/blob/b0686c8ccdbb8c74ab4f51bab708bc4f25160e29/src/SQLProvider/SqlDesignTime.fs#L171) as a proper 2D array, but [when it 'comes out' in `GetIndividual`](https://github.com/piaste/SQLProvider/blob/b0686c8ccdbb8c74ab4f51bab708bc4f25160e29/src/SQLProvider/SqlRuntime.DataContext.fs#L173) it's been flattened into a one-dimensional array, and the `SELECT` query fails. I'm not sure how to look further into it; I'm asking for any suggestions.

This PR required some refactoring, notably [replacing *in toto* the query that fetches column information](https://github.com/piaste/SQLProvider/blob/cb633642dff06c712e63c3a4c93bebe806666924/src/SQLProvider/Providers.Postgresql.fs#L671-L694), but the `pg_xxxx` tables have existed for a long time so that shouldn't cause compatibility issues.

Tested on PostgreSQL 9.6 and 10 beta 2.

I rebased on the latest master, and the commit history is a bit messy, so a squash and merge would probably work best.


piaste:
The CI failure seems to come from the SQLite-based tests. I don't think it has anything to do with the PR?

 
> C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1820,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "System.Data.SQLite.Linq". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors. 
 

> C:\projects\sqlprovider\tests\SqlProvider.Tests\CrudTests.fs(110,40): error FS0039: The field, constructor or member 'Create(Image)' is not defined [C:\projects\sqlprovider\tests\SqlProvider.Tests\SqlProvider.Tests.fsproj]

Thorium:
I tried to checkout this PR... The test failure does come from this PR. It can be reproduced by just trying to build SqlProvider.Tests.sln and in master it's working fine.

The ``Create(...)``` method is auto-generated from the table columns. Id-columns are skipped, if they are auto-increment, so the user don't need to input them. 

Interestingly enough, the test database `tests/SqlProvider.Tests/db/northwindEF.db` table `Pictures` column `Id` is missing the `key` and `autoincrement` specifications. So I think ``Create(Id, Image)`` is more correct in this case than ``Create(Image)``. So I think the test that is failing is invalid. So instead of

```fsharp
let savedEntity = dc.Main.Pictures.``Create(Image)`` imageBytes
```
...there should be...

```fsharp
let savedEntity = dc.Main.Pictures.``Create(Id, Image)``(1L, imageBytes)
```

But the problem with this PR is that all the other tables (like `Order`) have the Id field also included in the Create-column list, even if it should be autogenerated.

```fsharp
let x = dc.Main.Orders.``Create(OrderID)``(1L)
```

The user shouldn't need to input an Id here. The database should create the id.

Thorium:
There seems to be good modifications to make the code more clear. But I think the problem is in `src/SQLProvider/SqlDesignTime.fs` here:

Original:

```fsharp
|>  List.partition(fun c->c.IsNullable || c.IsPrimaryKey)
```

This PR:

```fsharp
|> Seq.filter (fun c-> (not c.IsNullable) || c.IsPrimaryKey)
```

See the `not` is applying the IsNullable, but not IsPrimaryKey. 
The definition in the database is `OrderId INTEGER NOT NULL PROMARY KEY AUTOINCREMENT`.

piaste:
Aaah, you're right. 🤦‍♂️ 

A while ago, I had seen [this issue](https://github.com/fsprojects/SQLProvider/issues/378) and I thought I could potentially fix it by detecting if the column was an autoincrement one (see the `is_sequence` column in the new query).

But then I realized it would be a breaking change, and it would belong in a different PR anyway, so I dropped it. I missed the change to the `requiredColumns` condition though.

Will update and fix.

Thorium:
I think breaking change is ok if it fixes a problem which we clearly have.

Thorium:
Merged, Thanks!

piaste:
I had the wrong number format for the `Pictures.Id` field when you merged. (Sorry, I didn't want to change all the local connections again and I was waiting for my own CI to check out.)

CI has now completed succesfully, you can merge from another PR [here](https://github.com/fsprojects/SQLProvider/pull/452).

pezipink:
great PR, thanks a lot!

PR #451, opened 2017-09-22
New ProvidedTypesContext.fs API from FSharp.TypeProviders.StarterPack

Thorium:
Supports cross-targeting type provider.


Thorium:
This will not add .net standard support yet, but is a step to that direction.

enricosada:
@Thorium make it crosstargeting as @dsyme suggestion, i can help doing the netstandard/infra/etc if you dont want to  👍  (ping me for any issue about netstandard/core/sdk, if you want)

Thorium:
I'm quite far already.
The branch is: https://github.com/fsprojects/SQLProvider/tree/netstandard

Needed code modifications on SQLProvider side should be already ok.

Current modifications needed: ProvidedTypes.fs needs a type hint at line 94:

```fsharp
|> Seq.forall (fun (originalName:string, newName:string) ->
```

(PR: https://github.com/fsprojects/FSharp.TypeProviders.StarterPack/pull/130)

.NET Standard project is located at `/src/SQLProvider.Standard/SQLProvider.Standard.fsproj`

The target frameworks are defined in the project file: `<TargetFrameworks>net461;netcoreapp2.0;netstandard2.0</TargetFrameworks>`
Corresponding builded dll files goes to root bin paths, e.g.: \bin\netcoreapp2.0

* Does build ok: netcoreapp2.0. Not working at runtime. 
* Doesn't build ok: netstandard2.0

The reason is System.Runtime reference problem. TypedReference used by FieldInfo: https://github.com/Microsoft/visualfsharp/issues/3303#issuecomment-329983262

enricosada:
thx for the info and the branch @Thorium . i'll try that, really curious.

afaik the project should be a .net core 2 fsproj (so without fsharp.net.sdk bebcause is deprecated, ref https://github.com/dotnet/netcorecli-fsc/wiki/How-to-migrate-1.0-projects-to-2.0 )
This is better also because the FscToolPath need to be changed to erasing TP to use the fsc .net full, and fsharp.net.sdk reference the fsharp.compiler.tools (so override the compiler to use probably)

Thorium:
Thanks. Now it builds for .NET Standard but at runtime (which is build for the user program) it says:

```fsharp
Could not load file or assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.
```

Thorium:
Ok, I think I got the branch working.

Branch https://github.com/fsprojects/SQLProvider/tree/netstandard is released as 1.1.12-alpha5 in the NuGet.org. More info: http://fsprojects.github.io/SQLProvider/core/netstandard.html

Issue #431, opened 2017-06-28
Firebird can't be used with case sensitive (quoted) table names


jlogar:
### Description

I have a DB that uses quoted (camel case) table names. I couldn't find a way to make the provider use correct names so all my queries fail.

### Repro steps

Do a query on a table that was created using quotes. Something like `"Assistant"`. Query using sth like:
```fsharp
    let ctx = sql.GetDataContext()
    let a = query {
             for ass in ctx.Dbo.Assistant do
             select ass
             }
```

### Expected behavior

Rows should be returned.

### Actual behavior

An exception occurs because the provider tries to use an unquoted table name so upper case is used by Firebird (eg. `ASSISTANT`). The query fails with:
```
SQL error code = -204
Table unknown
ASSISTANT
```
### Related information 

* Firebird
* Win 10
* .NET 4.6

jlogar:
taking a peek at the code, it would probably make sense to replace occurrences of:
`Table.Name.Replace("[","\"").Replace("]","\"").Replace("``","\"")`
with a separate function that would optionally (SqlDataProvider static parameter?) quote the table name.

thanks to @emmanueltouzery for pointing this out

Thorium:
There is a case sensitivity static parameter that may or may not help.

ODBC code has an option to change the quote char, but that is currently the only one.

pezipink:
would suggest we simply re-use the existing static parameter, shame it is named `OdbcQuoteCharacter` ...

Thorium:
Are you stating that Firebird is having multiple different quote chars? Or that the current one used is wrong?

pezipink:
ping @gibranrosa

jlogar:
I didn't notice the param in the docs before. Checked it out (http://fsprojects.github.io/SQLProvider/core/odbc.html) and it seems appropriate (apart from the naming :)).
Would the approach I mention above and using the `OdbcQuoteCharacter` be the way to go? I am really new to F# so might take me a while...

gibranrosa:
Hi! Sorry for the long delay! :/
@Thorium I haven't considered the case for quoted identifiers, but it is the double quote the correct one.
@jlogar This approach, I think, is the way to go, and be sure to test the case for Unquoted identifiers after the modification, which would be the OdbcQuoteCharacter being null, I suggest.

jlogar:
OK, so I finally got around to this again... The first thing I'm trying to do is to run Firebird specific tests so that I can add a test for the setting as the next step.
[...]

PR #453, opened 2017-09-25
Use double quotes for table names when enabled

jlogar:
`CamelCase` table with a single row was added. And a function to handle table name quoting. Also seems like a whitespace change creeped in in the random queries part, sorry about that.
BTW: Don't really like the way a live DB is committed into the sources...


gibranrosa:
Cool!

jlogar:
(related to issue #431)

Thorium:
Thanks!

Issue #396, opened 2017-04-21
Can't join a DB table with a result of another query


ArtemyB:
### Description

Can't join oa DB table with an `IQueryable` produced by another query. As a result of this I get `System.Exception: Unexpected join destination entity expression`.

**P.S.**: Seems like a problem exists even while joining only query results (not only query result + DB-table).

### Repro steps

Just a very simple example:
```
let q1 =
    query {
        for x in ctx.Dbo.TblXs do
        select x
    }
let q2 =
    query {
        for sample in ctx.Dbo.TblSamples do
        join x in q1 on (sample.Id = x.SampleId)
        select sample
    }
```

### Expected behavior

A normal query output(`IQueryable`) as it is for one big query directly to the DB-tables (not to another queries' resulting `IQueryable`s).

### Actual behavior

```
System.Exception: Unexpected join destination entity expression (FSharp.Data.Sql.Common.SqlExp+FilterClause).
   at Microsoft.FSharp.Linq.RuntimeHelpers.LeafExpressionConverter.EvaluateQuotation(FSharpExpr e)
   at Microsoft.FSharp.Linq.QueryModule.EvalNonNestedInner(CanEliminate canElim, FSharpExpr queryProducingSequence)
   at Microsoft.FSharp.Linq.QueryModule.clo@1730-1.Microsoft-FSharp-Linq-ForwardDeclarations-IQueryMethods-Execute[a,b](FSharpExpr`1 )
   at <StartupCode$FSI_0023>.$FSI_0023.main@()
Stopped due to error
```
### Known workarounds

Don't split the query on multiple smaller queries to compose them later (what is not always convenient).

### Related information 

* Microsoft SQL Server 2014
* Windows 10 Home, 64-bit
* .NET 4.5.2, F# 4.1

Thorium:
Currently not supported, as we support only one select-lambda... How complex is your real use-case?

Thorium:
However you can do Sysrem.Linq .Contains-operation to another IQueryable, resulting to a nested query.

ArtemyB:
@Thorium -

> How complex is your real use-case?

Not very. But an ability to have queries-to-queries is a helpful feature. I use it quite a lot im my case. Plus I also use  `groupBy`-expression, which is now also an issue for SQLProvider as I discovered. However as I'm working with SQL Server I can use FSharp.Data.Typeproviders, which solves my problem for now.

Thanks for the answer! 🙂

Thorium:
[...]

Issue #235, opened 2016-04-01
leftOuterJoin causing "unrecognized method call" with Postgresql


jackmott:
### Environment:
- Windows 10
- PostgresSQL 9.5
- .NET 4.6.1
- Npgsql 3.0.5 
- SQLProvider 0.0.11-alpha
### Query:

``` f#
 query {
            for timing in db.Public.Timings do
            leftOuterJoin sensor in db.Public.Sensors on (timing.Transpondercode = sensor.Sensorcode) into r1              
            select timing
        }        
```
### Error:

```
   System.Exception: unrecognised method call
    at Microsoft.FSharp.Linq.RuntimeHelpers.LeafExpressionConverter.EvaluateQuotation(FSharpExpr e)
    at Microsoft.FSharp.Linq.QueryModule.EvalNonNestedInner(CanEliminate canElim, FSharpExpr queryProducingSequence)   
    at Microsoft.FSharp.Linq.QueryModule.clo@1735-1.Microsoft-FSharp-Linq-ForwardDeclarations-IQueryMethods-Execute[a,b](FSharpExpr`1 )
    at AlphaFront.Db.getTimings(Double startTime, Double endTime) in C:\projects\AlphaFront\AlphaFront\Db.fs:line 488
```

This occurs whether I call `DefaultIfEmpty()` on `r1` and `r2` or not, and it occurs if I simplify the `select` clause to just return `timing`

Queries without left joins have been working great.

jackmott:
It appears I have figured out, via mining other issues, that leftOuterJoin is just not supported, except by use of the `!!` operator, maybe?  Some more details on how that works would be great.

pezipink:
Hello, yes we simply do not support `leftOuterJoin` yet because it also requires 'groupJoin' to be implemented and it has not been done.   I put the `!!` operator in as it was a cheeky way I could easily support left joins for things with properly defined relationships, without having to implement `groupJoin` and friends.

In theory it will be supported one day.

jackmott:
What is an order of magnitude on the effort involved in adding support?  If I take this up am I looking at weeks, months, or a deep deep rabbit hole? =)

pezipink:
It's a pretty tough one!  You would be basically be implementing all of aggregation with a bonus of having left joins along the way ...

Thorium:
You can now use `(!!)` also with traditional joins to convert them to left joins.
Not perfect (as you could want to join x > y condition), but works for the initial problem.

Issue #133, opened 2015-04-15
No sql is generated for skip clauses, in any provider.


toburger:
I don't have tested the other providers, but MSSQLSERVER and MSACCESS seem to ignore the Skip command.

colinbull:
It seems skip isn't not implemented on any of the providers. This should be an easy enough fix I'm hoping to get to this at the weekend.

pezipink:
I didn't do this originally for sql server, it is easy to do in 2012 but you have to mess around a bit in earlier versions using row number.  Completely forgot about it until now :)  It should be throwing there really ..

toburger:
Yes, I think it is advisable to throw an exception, if a query feature isn't implemented.

Thorium:
There are some other operators missing from the F# Query Expressions API as well:
https://msdn.microsoft.com/en-us/library/hh225374.aspx

E.g. headOrDefault is not working, but head will return the result that I would expect as headOrDefault (in the NuGet version). That is ok, I don't mind, it would just be nice to support the whole API.

pezipink:
The whole query expression API is big and we have several large, complicated bits missing, such as nested queries and aggregation (and indeed left join outside the my !! operator requires group join to be implemented)  - these are big projects.  Certainly more operators like headOrDefault are fairly simple to implement, we have a bunch of them already. 

The original non-erasing sql provider is simply a wrapper around sqlmetal so you get the whole linq expression tree for free :) 

It is also worth noting that in some cases, the examples there generate different sql to what we do, sometimes better and sometimes worse. 

Ross

colinbull:
FYI https://github.com/colinbull/SqlProvider/blob/feat_nicer_naming/tests/SqlProvider.Tests/QueryTests.fs this shows what is currently supported an not.

colinbull:
@pezipink This looks like it might be fixed? Is this the case?

pezipink:
I fixed it in the V1 branch for sql server 2012 and up...  mysql and postgres it was already working for I think.

Earlier versions of sql server this is a bit of a pain and hacky to achieve.

pezipink:
I lied, postgres is not implemented. MySql is though, and they both use `LIMIT` .. maybe postgres can use the same implementation? https://github.com/fsprojects/SQLProvider/blob/master/src/SQLProvider/Providers.MySql.fs#L531

colinbull:
[...]

PR #457, opened 2017-09-29
[WIP] Add SqlQueryBuilder wrapper

piaste:
I think it's really bad UX that the library lets you use unsupported `query` operators without any compile-time warning, and you have to look in the documentation to find out which ones are missing. People [keep](https://stackoverflow.com/questions/36360670/sqlprovider-query-errors-with-unrecognized-method-call) [being](https://github.com/fsprojects/SQLProvider/issues/133) [surprised](https://github.com/fsprojects/SQLProvider/issues/235) by that.

Although it would obviously be ideal to implement `groupJoin` and the other missing ones, I figured that there could be a simpler solution first: provide a `sqlQuery` computation expression where the unsupported operators are simply not available.

The `sql` builder in this PR is a dumb calque of [the original `query` CE](https://github.com/Microsoft/visualfsharp/blob/44fa027b308681a1b78a089e44fa1ab35ff77b41/src/fsharp/FSharp.Core/Query.fsi). The queries are still regular `IQueryables` so you can freely mix and match `sql` and `query` without issue (and it doesn't break existing code). As you can see, the tests that use unsupported operators become compile-time failures.

Other safety improvements that could potentially be achieved through this CE:

- I think it should be possible, and I would really like to, replace the `!!` operator with a more intuitive `naturalJoin` operator, like this:
```
for cust in dc.Main.Customers do
naturalJoin order in cust.``main.Orders by CustomerID`` do
select (order, cust)
```
  but I'm having trouble making it work - it doesn't seem like `CustomOperationAttribute` lets you define an operator that is "like `for`". Plus I'm not sure if the  `!!` operator would even be correctly recognized by the code in `SqlRuntime.Linq.fs` this way

- It might be possible to [ensure that `join` is only used on supported targets](https://github.com/fsprojects/SQLProvider/issues/396), by wrapping `QuerySource` in a SQL-specific wrapper class?

Thoughts?

### DIFF
--- a/src/SQLProvider/SqlProvider.fsproj
+++ b/src/SQLProvider/SqlProvider.fsproj
@@ -72,6 +72,7 @@
       <Link>ExpressionOptimizer.fs</Link>
     </Compile>
     <Compile Include="Operators.fs" />
+    <Compile Include="SqlQueryBuilder.fs" />
     <Compile Include="Utils.fs" />
     <Compile Include="SqlSchema.fs" />
     <Compile Include="DataTable.fs" />
--- a/src/SQLProvider/SqlQueryBuilder.fs
+++ b/src/SQLProvider/SqlQueryBuilder.fs
@@ -287,13 +287,21 @@ type SqlQueryBuilder() =
     /// </summary>
     [<CustomOperation("thenByNullableDescending",MaintainsVariableSpace=true,AllowIntoPattern=true)>] 
     member inline this.ThenByNullableDescending (source, [<ProjectionParameter>] keySelector) = this.QueryBuilder.ThenByNullableDescending (source, keySelector)
+      
+    //// WIP
+    
+    ///// <summary>
+    ///// A query operator that selects elements from another table based on a SQL foreign key relationship'. 
+    ///// </summary>
+    //[<CustomOperation("naturalJoin")>] 
+    //member inline this.NaturalJoin (source:QuerySource<'T,'Q>, body: 'T -> QuerySource<'Result,'Q2>) : QuerySource<'Result,'Q> = 
+    //  // QueryBuilder.For is defined as :
+    //  // QuerySource (Seq.collect (fun x -> (body x).Source) source.Source)
+    
+    //  // QueryBuilder.Join is defined as :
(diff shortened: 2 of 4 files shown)


Thorium:
I know I'm minority in this case but I think LINQ-query wrapper is ok, and we should implement the whole LINQ support rather than limiting ourselves to some custom developer interface which is not familiar to anyone. Most of the developers can do either SQL or LINQ. Actually the API for developers may be better as SQL like https://github.com/rspeele/Rezoom.SQL has done it.

And I think it's a good thing that we have multiple choices for a data access inside FSharp:
http://fsprojects.github.io/FSharp.Data.SqlClient/comparison.html

Thorium:
That being said if this is just a facade with good integration points, it could be nice addition. One thing to consider in general is how complex nested queries we want to support.

The current query didn't let me do countAsync etc. extensions. So they are now at Seq.headAsync etc, not an optimal place.

piaste:
Yes, it is as pure a wrapper around the LINQ query builder as one can write (in fact, I need to add a MIT attribution notice since I literally started by regex-replacing over `Query.fsi`). All the calls have the same name, parameters, and XML comments as the LINQ ones, and are forwarded inline straight to `query` without any business code at all. The only difference, so far, is that unsupported methods are omitted.

In fact, if the compatibility with LINQ ever gets to 100%, it would be possible then to simply delete all of this file and replace it with `let sql = QueryBuilder()`. Any additional operators could become extension methods to QueryBuilder. (Speaking of which, do you remember which kind of problems you ran into with adding `countAsync`? [This seems to work](https://dotnetfiddle.net/Uo6tTr).)

And, assuming it's doable in the first place, I can see the case against adding any SQL-specific operators such as `naturalJoin`, in order to keep things as clear as possible. Thing is, currently people have to learn about the `!!` operator and how to use it to get a left-join, so I think it would be an improvement if I could get that operator to at least to pop up in Intellisense when you type `join`. Possible counter-argument: it's implemented in a way that's totally different from the regular LINQ operations, so it *should* appear different. I'm not sure.

Thorium:
https://github.com/fsharp/fslang-suggestions/issues/495
Yes it would work with custom query, but not with F# default.

PR #467, opened 2017-10-26
[WIP] Fix for procedure return parameters

Thorium:
Ok, this is working already. You can execute the stored procedures from the test project.

But the problem is now that all the procedure return-types are in the root namespace FSharp.Data.Sql
The problem are lines 291 & 312 of SqlDesignTime.fs. 

I would need a little help form some stored-procedure-guru... 
@colinbull would you have extra 20 minutes to look at this?


Thorium:
This is better than non-working sprocs, I'll merge already.

Issue #470, opened 2017-10-26
Duplicate sproc names ("GetMethodImpl not support overloads")


piaste:
See https://github.com/fsprojects/SQLProvider/issues/465#issuecomment-339382409 and followup comments

I noticed the same 'not support overloads' issue with PostgreSQL when accessing the `Functions` property.

The problem in that case was that while the provider makes sure not to fetch sprocs with the same name, the design time then applies `nicePascalName` and it may result in a conflict if two sprocs only differ by non-alphanumeric characters.

Now if the functions are user-created you could just name them more sensibly, but in my case they came from a Postgres extension so I couldn't do anything about it. 

I have [a fix for the issue](https://github.com/fsprojects/SQLProvider/compare/master...piaste:fix_pg_extension_functions) as it relates to Postgres and I guess it's technically submittable, but I realized it's a super hacky one because it just groups by `nicePascalName` upstream in `GetSprocs` (and only in the Postgres provider).

The proper fix would be to handle pascal-duplicated names in SqlDesignTime, during the [tree-walking](https://github.com/fsprojects/SQLProvider/blob/master/src/SQLProvider/SqlDesignTime.fs#L345). I'll get to work on this, just wanted to check if my reasoning was sound.

(Also, ideally we would not simply hide overloaded (or pseudo-overloaded) procedures, we ought instead to give them proper F# names, by adding ```` or something. Not a big deal, though, as it's easy enough to add a wrapper procedure with a different name if necessary.)

Thorium:
Maybe a concurrent dictionary just after and before pascal-naming to check duplicates, and a tooltip to display original name?

Thorium:
Sorry to change SqlDesignTime.fs so often...

piaste:
's ok, I needed some practice with rebase anyway 🙌 

With the current master the sprocs don't crash outright, so it's much less of an issue, but it's still kinda buggy - only one shows up, and the XML comments are all merged. Submitted a PR.

Thorium:
Thanks, PR merged to version 1.1.19

PR #476, opened 2017-10-30
Fix sproc name overload + Fix Postgresql sprocs

piaste:
General fix:
 - If multiple stored procedures would get the same .NET names, the additional ones get apostrophes added to make them unique - eg. `Foo`, `Foo'`. See #470 

Postgres fixes:
 - No longer shows sprocs that the user isn't allowed to execute
 - No longer shows trigger sprocs (which cannot be manually executed)
 - No longer shows sprocs where any required parameter (input or input-output) lacks a suitable type mapping
 - Anonymous parameters in stored procedures are now given generic `param1`, `param2`, etc. names on the .NET side. Doesn't seem to break anything, even in the unlikely case where you explicitly named your *second* parameter `param1` for some reason.

Unrelated:
 - Ran the latest Paket and it updated Paket.Restore.Targets


Thorium:
Thanks!

Issue #479, opened 2017-10-31
Oracle schema reader may pick up columns from wrong tables


object:
### Description

I our Oracle database we have several versions of the schema registered under different owners. The SQL Provider schema reader collects all columns with the same table name, no matter where they belong. This causes exception later.

### Repro steps

1. Create a database with table "Test", few columns and owner John
2. Add the database with the same schema for owner Bill, but expand the table "Test" with one extra column "Extra"
3. Try retrieving data from John.Test (don't specify individual columns, select everything)
4. The query fails with Oracle exceptions stating that "Extra" column is invalid

### Expected behavior

Tables columns should be populated according their owners.

### Actual behavior

Current implementation of the provider uses the following query to select columns:

select data_type, nullable, column_name, data_length from all_tab_columns where table_name = [table_name]

There may be several tables with the name [table_name]. The provider should take into account "owner" column value.

### Related information 

* Oracle 11
* Windows 10
* .NET 4.5.1

Thorium:
Thanks, PR merged to version1.1.19.

PR #480, opened 2017-11-01
Fix for #479 (column name resolution in Oracle).

object:
Tested on our Oracle database, table columns are correctly resolved now (the problem is described in #479 ).


Thorium:
Thanks!

Issue #483, opened 2017-11-03
Question about more idiomatic fsharp option check for exactlyOneOrDefault


giuliohome:
### Description

This is a follow up to issue [#478](https://github.com/fsprojects/SQLProvider/issues/478)

### Repro steps

Steps required to reproduce the problem

1. Step A
Do everything like for the above mentioned issue #478

2. Step B
Then try to write a more functional/idiomatic code so change this code

```fsharp
        let foundPrice = 
             query {
             for p in prices do 
             where (p.Provider = aPrice.Provider  && p.RefDate = Some aPrice.RefDate)
             exactlyOneOrDefault
             } 
         
        if not(isNull foundPrice) then
            foundPrice.MyValue <- Some (aPrice.Price + 1.1)
          else 
            let newRec = prices.Create()  
            newRec.RefDate <- Some aPrice.RefDate
            newRec.Provider <- aPrice.Provider
            newRec.MyValue <- Some aPrice.Price
```

to a more idiomatic

```fsharp
        let foundPriceMaybe = 
             query {
             for p in prices do 
             where (p.Provider = aPrice.Provider  && p.RefDate = Some aPrice.RefDate)
             select (Some p)
             exactlyOneOrDefault
             } 
         
        match foundPriceMaybe with
        |Some foundPrice ->
            foundPrice.MyValue <- Some (aPrice.Price + 1.1)
        |None ->
            let newRec = prices.Create()  
            newRec.RefDate <- Some aPrice.RefDate
            newRec.Provider <- aPrice.Provider
            newRec.MyValue <- Some aPrice.Price
```

### Expected behavior

No error, same behavior, just a nicer source code

### Actual behavior

I get a runtime exception when I submit the context updates  

```fsharp
System.Exception was unhandled
Message: An unhandled exception of type 'System.Exception' occurred in FSharp.Data.SqlProvider.dll
Additional information: Error - you cannot update an entity that does not have a primary key. (main.very_urgent)
```

### Known workarounds

I'm forced to stay with the ugly null check

### Related information 

Same as previous issue except of course latest package 1.1.19

gaston-e-nusimovich:
The null check, while not as elegant, is effective.

Thorium:
Meanwhile I use myself:
```fsharp
async {
   let! foundPriceMaybe = 
       query {
           for p in prices do 
           where (p.Provider = aPrice.Provider  && p.RefDate = Some aPrice.RefDate)
           select (p)
[...]

PR #484, opened 2017-11-03
REPRO test cases for #issue #483

giuliohome:
select (Some ...) only works when... is a tuple e.g. (cust,ord) from a join 
but it doesn't when it is a simple select (Some cust) from a table
in the latter case there are no meaningful keys/values in the data dictionary of the query builder


Thorium:
Now that I look the code it seems so obvious it was missing. ;-)
Thanks!

PR #491, opened 2017-11-09
complete unit test and fix for issue #483  

giuliohome:
complete unit test and fix for issue #483  
**update** selected record after pattern matching with option from exactlyOneOrDefault


Thorium:
Thanks!

PR #497, opened 2017-12-14
Clarify that ``delete all items from single table`` immediately stores changes

cmeeren:
Based on observed behavior.


Thorium:
Thanks!

PR #508, opened 2018-01-25
Fixes for Firebird's numeric column, dialect 1 tests and column's descriptions.

gibranrosa:
- Columns with Numeric type wasn't been generated
- Tests wasn't working with databases with Dialect 1.
- Column's descriptions wasn't been gathered.


Thorium:
Thanks!

Thorium:
We get each time new versions of two of these 1-2MB database files (*.FDB) what is unsustainable for the Git repository. Git will each time save the binary files to the history, increasing the repository size, and that cannot be removed. So after 500 commits, each new clone of the repository will take 1GB-2GB more of disk space.

Can we do something about that, still maintaining the test setup be reasonable?

PR #511, opened 2018-01-30
Support for group-by over a constant

Thorium:
I wanted to return multiple SQL operations with a single SQL:
```sql
SELECT SUM(Price), AVG(Price), SUM(SoldUnits) FROM Products
```
Sadly there is no easy way to do that in F# [query](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/query-expressions) and also C# EntityFramework solution is kind of a [hack](https://social.msdn.microsoft.com/Forums/en-US/0b595016-a8ca-41b5-a67a-6adc074b9f91/how-to-make-a-linq-query-return-multiple-count-or-sum-columns?forum=linqprojectgeneral): group by a constant value.

```fsharp
let qry =  
  query { 
      for p in dc.Main.Products do 
      groupBy 1 into g 
      select (g.Sum(fun p -> p.Price), g.Average(fun p -> p.Price), g.Sum(fun p -> p.SoldUnits)) 
  } |> Seq.head 
```

Well, that is not pretty, but after this PR it works. 
(Only a single table, and it won't generate the extra-group by as it wouldn't be valid SQL.)


Thorium:
Build is failing also in master, due to Postgres CI, https://travis-ci.org/fsprojects/SQLProvider/jobs/335392877

PR #513, opened 2018-01-31
Fix Travis delayed initialization

piaste:
Adds a loop to ensure that the `createuser` init command goes through even if it takes a few more split-seconds for the postgresql service to start.

As a test, I've restarted the job five times just to be sure, and hasn't failed so far (most time taken has been 200ms).


Thorium:
Thanks!

Issue #390, opened 2017-04-07
SQL Server String Operations


adam-becker:
This is more of a question than anything else. Are there any plans to support the SQL Server string operations - Left, Right, Substring, etc?

I have plenty of queries that use these, and the only workable solution at the moment is to create an SP that does the query and call that using this library. However I've run into a case where I'm not allowed to create an SP on the database I'm querying.

Example:

```fsharp
    select
      { Field1 = e.Id;
        Field2 =
          if (e.Id =% "[0-9]%_[0-9]%" && e.Id <>% "%_0[0-9]%") ||
             (e.Id =% "%_[0-9]%_0%")
          then Left( e.Id, 5 )
          else e.Id;
      }
```

I don't see any way to accomplish the same thing without these string operations.

Thorium:
Good question!

There would be string operations to different providers available. I'm not big fan of that. But what would you / our users think?

Pros:

* Get things done on some non-common use-cases
* To do this now you have to either duplicating data in database (to have partial strings in some separate column) or create a procedure

Cons:

* Will even more blur the line of .NET and SQL, what is supported, and what is executed and where.
* May cause queries not to hit database indexes, so the queries would be slow on large tables.
* We couldn't create here a compile-time verification anyways
* May not be standard, so the functionality may differ a bit on each provider.
* Solving string-parsing problems in 2017 is usually sign of poor domain modelling.

Related to #388

adam-becker:
I wouldn't mind seeing database specific operations exposed in their own modules. Projects like Xamarin attempt to create a flat API surface for all devices, but still expose device specific functionality if you want to use it.

When working with a database, you know which database you're accessing. It's akin to storing a connection string which are typically database specific, so I don't see any problems with this sort of thing. Something like this:

```fsharp
namespace FSharp.Data.Sql.IntrinsicFunctions

module SqlServer =
  let left str count = ...
  let right str count = ...
  let substring str start count = ...
```

Calling them intrinsics, and placing them inside database specific modules I think keeps the line between .NET and SQL in focus. The blurring tends to happen when existing .NET functions are automatically translated behind the scenes, which I dislike as well. If you saw the following code, you'd be able to tell that the code is a SQL function:

```fsharp
select
  { Field1 = SqlServer.left item.Column 5;
    Field2 = "";
  }
```

I think type safety is still possible too, the signature would look like:

```fsharp
val left : string -> int -> string
```

Unless you meant something else?

[...]

PR #515, opened 2018-02-07
WIP: Case when else support

Thorium:
Progressing slowly towards #390 ... Sorry that I had to re-order some functions again.
We could start with simple "if this is a column or columns or entity -> use the current functionality, else if this is a SQL-canonical-operation-tree, serialize that for the select-clause". Or maybe even more simple for a start.

* [x] Move filterBuilder functions to be allowed to use also on `case when filter then x else y`
* [x] Add parametrized SQL possibility to canonical operations
* [x] Parse IIF-ConditionExpressions from LINQ-trees
* [x] Add support SQL support for Case-when-select-clauses to where-clauses
* [x] Create SimpleCondition to case-pattern to replicate the current where Condition -active pattern 
* [x] Change projection columns from string arrays to strongly typed ProjectionParameter
* [ ] Populate ProjectionParameter canonical operation trees in ProjectionItem pattern of transform-function in SqlRuntime.QueryExpression
* [ ] Parse ProjectionParameter canonical operation trees in columns of select-clauses
* [ ] Ensure projection lambda parameter types work also for the new canonical oprations
* [x] Ensure all canonical operations are producing parametrized SQL
* [x] Merge SimpleCondition and Condition if possible
* [ ] Update code comments

So the case-when-else is already supported in where-clause, but not yet parsed as an option from projection lambdas

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,9 @@
+### 1.1.30 - 01.02.2018
+* Postgres ltree mapping PR #510
+* Single table group-by over a constant value PR #511
+* More aggregate functions: StdDev, Variance
+* More canonical math functions: Sqrt, Sin, Cos, Tan, ASin, ACos, ATan, Math.Max and Math.Min
+
 ### 1.1.29 - 25.01.2018
 * Firebird fixes for numeric columns and column descriptions, PR #508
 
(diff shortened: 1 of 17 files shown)


Thorium:
Where and aggregates like sum/max/min/avg can already use case-clause.
So only projection lambda refactoring is not done.
Maybe this could be merged already and modify the projection-lambda more as separate task.

PR #518, opened 2018-02-14
Case when else support in select #515

Thorium:
Continued on #515 

```fsharp
        query {
            for o in dc.Main.Orders do
            for c in dc.Main.Customers do
            where (o.ShipCity = c.City && c.Country.Contains("u"))
            select (
                (if c.City.StartsWith("Lon") then
                    "Greater London"
                 else
                    c.City),
                c.Country
            )
        } |> Seq.toArray
```

Produces now (this sample is in SQLite, the SQL varies based on database engine):
```sql
     SELECT
        CASE WHEN ([Customers].[City] LIKE @param1) THEN 
           @param2 
        ELSE [Customers].[City] END as [op1716912576],
        [Customers].[Country] as '[Customers].[Country]' 
    
     FROM main.Orders as [Orders] , 
          main.Customers as [Customers] 

     WHERE (([Orders].[ShipCity] = [Customers].[City] AND [Customers].[Country] LIKE @param4))

       -- params @param1 - "Lon%"; @param2 - "Greater London"; @param4 - "%u%";
```

So all the infrastructure is there now. The only thing that should be done, is to decide which part of projection-lambda-transform is converted to SQL-side and which is still used in BL side.

The current one uses OperationColumnOnly -active pattern to identify simple single-column-only select-clause.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+### 1.1.33 - 14.02.2018
+* Canonical operations chaining type-checking fixed
+
 ### 1.1.32 - 13.02.2018
 * initial case-when-else support on where-clauses, PR #515
 * mapping SQL server time to TimeSpan, PR #517
--- a/docs/content/core/parameters.fsx
+++ b/docs/content/core/parameters.fsx
@@ -141,5 +141,6 @@ Besides the static parameters the `.GetDataContext(...)` method has optional par
 * resolutionPath - The location to look for dynamically loaded assemblies containing database vendor specific connections and custom types
 * transactionOptions - TransactionOptions for the transaction created on SubmitChanges.
 * commandTimeout - SQL command timeout. Maximum time for single SQL-command in seconds.
+* selectOperations - Execute select-clause operations in SQL database rahter than .NET-side.
 			  
 *)
--- a/docs/content/core/querying.fsx
+++ b/docs/content/core/querying.fsx
@@ -97,6 +97,55 @@ let itemAsync =
 
 If you consider using asynchronous queries, read more from the [async documentation](async.html).
 
+## SELECT -clause operations
+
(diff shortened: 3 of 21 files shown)


Thorium:
This leads to quite interesting design decisions. There is already some discussion at #390.

Advantages of serializing operations to SQL-select-clause:

 + Smaller data-transfers, if you select e.g. Substring(4) instead of whole text, or you can use CASE to avoid unnecessary columns
 + Cross-join is a realistic option
 + May avoid queries to select items that are not indexed
 + In theory more control to the developer if (s)he has a patience (or skills) to write non-sql code outside the query.

Disadvantages of serializing operations to SQL-select-clause:

 - Some operations are not supported on all the SQL-engines.
 - Possible unexpected results for .NET operations, due to database doing things differently: 
   may have different results, e.g. roundings, substring index position, datetime calculations, ...
 - Multiple operations to the same columns could cause more data over the wire.
 - May cause (Oracle) to not being able to use database column indexes.
 - User's source code may not be as compact as now.
 - Possible SQL-errors are runtime failures, not compile-time.

So, should we by default:

- Use SQL-side select projections on all other than SqlEntities?
- Use .NET BL side select projections on all multi-column selects?
- Something else?
- Have more constructor parameters? But this is not a context-option I think, it should be a query-based...
- So should we have methods? Something like Seq.ToList but Seq.executeProjectionInSQL ? Or Seq.``execute Projection In SQL``()
- Should we move all SQLProvider seq based operations to a custom class? Instead of Seq.toList would be Sql.ToList ?

Thorium:
Ok, I'm actually quite happy with the result.

Thorium:
This would be ready to merge, but as ODBC-engines do vary the keyword support, there has to be option to disable this and use the old select-projection. Maybe to add one more parameter to GetDataContext(...)? Should the default be execute-select-operations-in-client or execute-select-operations-in-database?

PR #524, opened 2018-02-27
Add ConnectionStringHash property to SqlEventData

piaste:
This PR adds a `ConnectionString` property to the `SqlQueryEvent` payload.

It shouldn't be a breaking change, unless a user is for some reason creating his own instances of the `SqlQueryData` record.

It also shouldn't be a security issue, since anybody that can read the payload can already read all input parameters to all queries, which is much more sensitive information.

Adding this property allows us to figure out which connection generated a particular query. This can have several uses; in my case, [I have one connection on a tight polling loop and I want to exclude it from ordinary query logs](https://github.com/npgsql/npgsql/issues/1786).

(Note: I considered including the `Connection` object directly, instead. The upside is that it would provide more information, depending on the particular `IDbConnection` implementation in use, such as ongoing transactions, connector IDs, etc. - or you could simply compare the connection object by reference. The downside is that I can easily imagine a user doing all kinds of awful hacks like changing connection targets in mid-query, or just accidentally disposing the connection.)

### DIFF
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -53,10 +53,10 @@
     </PropertyGroup>
 
     <!-- If shasum and awk exist get the hashes -->
-    <Exec Condition=" '$(PaketRestoreCachedHasher)' != '' " Command="$(PaketRestoreCachedHasher)" ConsoleToMSBuild='true'>
+    <Exec StandardOutputImportance="Low" Condition=" '$(PaketRestoreCachedHasher)' != '' " Command="$(PaketRestoreCachedHasher)" ConsoleToMSBuild='true'>
         <Output TaskParameter="ConsoleOutput" PropertyName="PaketRestoreCachedHash" />
     </Exec>
-    <Exec Condition=" '$(PaketRestoreLockFileHasher)' != '' " Command="$(PaketRestoreLockFileHasher)" ConsoleToMSBuild='true'>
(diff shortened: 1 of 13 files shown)


Thorium:
Usually runtime connection string is crypted in the web-server. I do argue that "all input parameters to all queries" is not as sensitive information. 

Even though your users would have a read-access to the database, this does expose a database system user credentials that can alter the data.

Thorium:
Can't we just add a hash-code of the connection string? That you could compare to whatever you want, to not mess multiple servers.

piaste:
> Can't we just add only a hash-code of the connection string? That you could compare to whatever you want, to not mess multiple servers.

Yes, that works for me. I've revised the PR.

I've used a proper crypto hash rather than a quick & simple `GetHashCode()` because:

 1. While SqlProvider wasn't using `System.Security.Cryptography` before, the library is part of .NET Standard so it doesn't create additional dependencies
 2. You raised a security concern, so using an unsafe hash didn't seem appropriate
 3. `GetHashCode()` [tends to collide even for small sample sizes](http://kenneththorman.blogspot.it/2010/09/c-net-equals-and-gethashcode.html)
 4. I'm caching the result so there shouldn't be measurable performance costs

Incidentally, I've noticed that the helper methods for raising query events (`PublishSqlQuery`, `PublishSqlQueryCol`, `PublishSqlQueryICol` and `PublishExpression`) are publicly exposed. Shouldn't they be internal?

Thorium:
I think `SqlQueryEvent` is the only event that has to be public that users know what the SQLProvider actually does. The other events are not so important, they could be internal.

Thorium:
Thanks!

PR #530, opened 2018-03-27
Allow specifying multiple schemas for PostgreSQL

piaste:
When connecting to a PostgreSQL database, the `owner` parameter is used to specify the schema, by inserting it as a `WHERE` clause in the `GetTables` method.

This simple PR lets you specify multiple schemas so you can access them under a single provider type. To maintain backwards compatibility, the schemas must be specified in the string, separated by semicolons, commas, newlines, and/or spaces.

(I've noticed that the Travis builds have been failing since, apparently, the latest `paket update`. If I can figure out the problem I'll submit a separate PR.)


Thorium:
Great. This should be quite simple to add also for MySQL which is using the owner parameter also.

PR #531, opened 2018-04-01
Show how to use connection string from App.config.

brettrowberry:


Thorium:
More documentation never hurts, I'll merge this and maybe we can add some note later. Thanks.

PR #537, opened 2018-04-13
Added support for TIMESTAMP with precision specification.

object:
I've added support for Oracle TIMESTAMP with precision specification.
TIMESTAMP columns may include fractional seconds precision specification, e.g.:

TIMESTAMP(6)
TIMESTAMP(3) WITH TIMEZONE
TIMESTAMP(6) WITH LOCAL TIMEZONE

Precision must be stripped from the column type string, otherwise provider won't find the type.


object:
The build failed but this is not related to my PR. The error was "Could not locate the assembly "System.Data.SQLite.Linq" which has nothing to do with the change.

Thorium:
Thanks.

PR #544, opened 2018-05-03
Improved option type support in convertTypes.

object:
If option types are enabled (UseOptionTypes is set to true), then SQLProvider fails to retrieve results in some aggregate queries, here are examples:

1. Create the following table
create table test_a (
    a_id_not_null number(5,0) not null,
    a_text_not_null varchar2(20) not null,
    a_id_null number(5,0) null,
    a_text_null varchar2(20) null)

2. Insert test data
insert into test_a (a_id_not_null, a_text_not_null, a_id_null, a_text_null) values (45, 'TEXT_VALUE', 45, 'TEXT_VALUE');

3. The following queries fail with conversion exception (inside convertType function):

query { for t in ctx.TestA do
groupBy t.AIdNotNull into g
select (g.Min(fun p -> p.ATextNull)) }

query { for t in ctx.TestA do
groupBy t.AIdNull into g
select (g.Min(fun p -> p.ATextNotNull)) }

There are a few reasons for failures:

1. convertType test that option types have a name prefixed with "Option". Not sure whether this was the case in earlier versions of F#, but present option type use the name "FSharpOption". I kept the test for both, just in case.

2. When instantiating an option type, convertType used a call to unbox which resulted in Option<Object>, since at the time of unboxing compiler couldn't infer the type of underlying value. This caused exception in Convert.ChangeType which couldn't convert between Option<Object> and a primitive type. I had to replace unbox with specific type checks.

3. Missing conversions for certain types, including String.

Current PR fixes these problems, the above queries work.

### DIFF
--- a/src/SQLProvider/Utils.fs
+++ b/src/SQLProvider/Utils.fs
@@ -68,35 +68,66 @@ module internal Utilities =
         parseAggregates' fieldNotat fieldNotationAlias query []
 
     let rec convertTypes (itm:obj) (returnType:Type) =
-        if returnType.Name.StartsWith("Option") && returnType.GenericTypeArguments.Length = 1 then
+        if (returnType.Name.StartsWith("Option") || returnType.Name.StartsWith("FSharpOption")) && returnType.GenericTypeArguments.Length = 1 then
             if itm = null then None |> box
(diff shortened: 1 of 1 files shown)


Thorium:
Thanks! 
I've always checked option types as `.IsNone` and then `.Value` but haven't tried those a lot with group-by.

PR #545, opened 2018-05-03
Context schema cache support

object:
I ran more tests (applied your PR), and it worked great. I was able to generate a schema cache for the whole solution, then disconnect from the database and use the cache in compilation. The issue I had with incomplete cache was no longer reproducible.


Thorium:
Great! Now we would need documentation, but the documentation generation is also broken at least in my machine.

object:
Documentation generation is also broken on my machine for the moment. I can write a doc if you point me to right file/section where is should be written.

Thorium:
It could be its own page, but this should also do:
https://github.com/fsprojects/SQLProvider/blob/master/docs/content/core/parameters.fsx

object:
Sent a PR with updated docs.

Issue #238, opened 2016-04-07
Multi-threading issue on server side.


Thorium:
I'm running on this issue when I have a web-server that uses SQLProvider and there are multiple requests at the same time (from different users, or from async JavaScript). e.g.:

- User A calls `let e = query { for m in dbContext.MyTable do where (m.Id = "A") select (m) }`
- User B calls `let e = query { for m in dbContext.MyTable do where (m.Id = "B") select (m) }`
- User A continues with `e.SetColumn("mycol", 1)` and `dbContext.SubmitUpdates()`
- User B continues with `e.SetColumn("mycol", 2)` and `dbContext.SubmitUpdates()`

Crash `Unchanged entity encountered in update list - this should not be possible!`

The problem is basically the same as #173 but how should I do a good workaround?
These users are not touching the same rows in the database.

pezipink:
I am not 100% sure why this is happening - if it is a sync problem can you not just lock or sync via a mailbox?

I will look at this when I get back around to doing some work on this type provider.

Thorium:
Are you suggesting that I would have only one web-user accessing the database at time?

I would rather like to see some kind of "if this error, refresh the context and retry"-operation...

pezipink:
You could have multiple data contexts, there is no real disadvantage to this since the connections are closed and reopened with each action, which are pooled via normal ado mechanics - would that sort it ?

Thorium:
Ok I think this is a basic problem that gets solved if I wrap my queries to transactions so that both `query` and the `SubmitUpdates` are inside of a same transaction. By default, they are not; how could query know anything.

So basically if I have read-only-operations they are ok to execute as is, but if I have SubmitUpdates then I have to wrap the queries related to this to a transaction. Inside SubmitUpdates is made a new transaction with default options (which is enlist to existing transaction or make a new if none exists) which is good.

I made a little wrapper to do the transaction:
```

let dbReadOnlyContext = 
    if cstr = null then TypeProviderConnection.GetDataContext()
    else TypeProviderConnection.GetDataContext cstr

let writeWithDbContextManualComplete<'T>() =
    let isMono = Type.GetType ("Mono.Runtime") <> null
    let context = TypeProviderConnection.GetDataContext cstr
    let scope =
        match isMono with
        | true -> new Transactions.TransactionScope()
        | false ->
            // Mono would fail to compilation, so we have to construct this via reflection:
[...]

PR #548, opened 2018-05-08
Upsert support for SQLite and PostgreSQL

piaste:
I know I've got some other PRs to finish, but this one was rather straightforward and it would significantly improve (some parts of) my project, by removing the fake upserts I currently do with if-select-then-update-else-insert (which are slow and vulnerable to race conditions).

You just set `myRow.OnConflict <- OnConflict.Update` or `myRow.OnConflict <- OnConflict.DoNothing` and, if the provider supports it, the INSERT query will be handled appropriately by either updating the existing record with the new record's values, or by ignoring the conflict without throwing an exception. Not much to it.

UX consideration: currently the `OnConflict` type is conservatively placed the `FSharp.Data.Sql.Common` namespace. Since its use could be somewhat pervasive, might there be a better place? I think the only namespace that currently gets imported frequently is `FSharp.Data.Sql.Operators`, where it doesn't fit.

For other providers: SQL Server and Oracle have upsert support in the form of the MERGE statement, but they're a little trickier to format, so I intend to try them it after I get the CI integration working for it. Also, they apparently aren't 100% safe against race conditions, so I'm not quite clear on what they offer over a classic "UPDATE.... IF @@ROWCOUNT = 0 THEN INSERT".

### DIFF
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -18,6 +18,14 @@
     <PaketExePath Condition=" '$(PaketExePath)' == '' ">$(PaketToolsPath)paket.exe</PaketExePath>
     <PaketCommand Condition=" '$(OS)' == 'Windows_NT'">"$(PaketExePath)"</PaketCommand>
     <PaketCommand Condition=" '$(OS)' != 'Windows_NT' ">$(MonoPath) --runtime=v4.0.30319 "$(PaketExePath)"</PaketCommand>
+
+    <!-- .net core fdd -->
+    <_PaketExeExtension>$([System.IO.Path]::GetExtension("$(PaketExePath)"))</_PaketExeExtension>
+    <PaketCommand Condition=" '$(_PaketExeExtension)' == '.dll' ">dotnet "$(PaketExePath)"</PaketCommand>
+
+    <!-- no extension is a shell script -->
+    <PaketCommand Condition=" '$(_PaketExeExtension)' == '' ">"$(PaketExePath)"</PaketCommand>
+
     <PaketBootStrapperExePath Condition=" '$(PaketBootStrapperExePath)' == '' AND Exists('$(PaketRootPath)paket.bootstrapper.exe')">$(PaketRootPath)paket.bootstrapper.exe</PaketBootStrapperExePath>
     <PaketBootStrapperExePath Condition=" '$(PaketBootStrapperExePath)' == '' ">$(PaketToolsPath)paket.bootstrapper.exe</PaketBootStrapperExePath>
     <PaketBootStrapperCommand Condition=" '$(OS)' == 'Windows_NT'">"$(PaketBootStrapperExePath)"</PaketBootStrapperCommand>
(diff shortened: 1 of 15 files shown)


Thorium:
I think if you wrap your code to a System.Transaction -transaction, we commit to that, and no conflicts will ever happen. Am I wrong?

Thorium:
No, I meant that the select-query and the update should be inside the same transaction:

There should never be conflicts no matter how many items you run in parallel with this kind of code:

```fsharp
let createOrUpdate x =
   use scope = new Transactions.TransactionScope(Transactions.TransactionScopeAsyncFlowOption.Enabled)
   let ctx = TypeProviderConnection.GetDataContext cstr
   let myItem = query { for i in ctx.MyTble where i.Id=x select i} |> Seq.tryHead
   match myItem with
   | Some x -> x.Thing <- "updated"
   | None -> ctx.MyTble.``Create(...)``(...)
   ctx.SubmitUpdates()
   if scope <> null then 
      scope.Complete()
   myItem
```

...because the transaction by definition should should span over the whole function call causing the table to be locking the item so that no-one is able to update it at the same time.

I had similar issues before, see: https://github.com/fsprojects/SQLProvider/issues/238

piaste:
Ok, I think I get what you meant now. Unfortunately that approach doesn't cover all cases.

Transactions can ensure that UPDATEs and DELETEs are executed sequentially and play nice with each other. But that only works because you run the same command whether it's the first or the second in the execution order.

With INSERTs (which this PR targets) you need to run either INSERT or UPDATE based on who gets the write lock first. What happens is that, even wrapped in different transactions, multiple SELECTs will be allowed to execute concurrently, and each one will find that the key doesn't exist. *Then* each transaction will attempt an INSERT, and at that point they will acquire write locks and run in order, but it's too late - they have already decided to run an INSERT, and will fail.

To work properly, either (a) the database would need to acquire full table-locks on every SELECT, which no database does for obvious reasons; or (b) the provider would need to emit a single query for both operations (CASE WHEN NOT EXISTS(..) THEN INSERT (..) ELSE X (..)), which would acquire a write lock immediately, but that's just a more complicated way to do the ON CONFLICT X clause.

I tested it to be sure - if I run the following code:

```f#
[ for i in 1 .. 20 do
    let u = username + (string i)        
    yield async {
      use scope = new Transactions.TransactionScope(
        Transactions.TransactionScopeAsyncFlowOption.Enabled
      )
    
      let ctx = Azienda.GetDataContext connstr
      let myItem = 
        query { 
          for l in ctx.Public.Login do
          where (l.loginid = loginid)
          select l
        } |> Seq.tryHead
    
      match myItem with
      | Some x -> x.Username <- u
      | None ->
        let login = ctx.Public.Login.``Create(email, password_hash, password_salt, username)``
                                             (email, [||], salt, u)
    
        login.loginid <- loginid
        login.Username <- u
        //login.OnConflict <- FSharp.Data.Sql.Common.OnConflict.Update
    
      ctx.SubmitUpdates()
    }
]
|> Async.Parallel
|> Async.RunSynchronously

```

then, if I use an already existing `loginid`, all the updates run smoothly (and the final result is either `username19` or `username20`, at random). But if I use a `loginid` that doesn't already exist, I get the unique constraint violation:

```
 {Npgsql.PostgresException (0x80004005): 23505: duplicate key value violates unique constraint "login_loginid_key"
```

If I uncomment the `OnConflict` property, then the inserts work, even if I comment out the transaction line.

---

One more thing: while I do have a real use case for the particular scenario above (parallel jobs importing from partially-overlapping data sources),  I think the improved usability, reduced boilerplate and improved performance would be valuable on their own, for the vastly more common case where you're just doing a regular upsert.

I know that speed isn't the focus of this library (and it shouldn't be), but boring ETL jobs where missing a column is easy is exactly where I appreciate SQLProvider's type-checking the most. And adding SELECT checks to properly handle concurrent imports slowed down those jobs by roughly half, and made the code look a fair bit worse.

Fixing that by using the databases' native tools, and getting rid of some ugly support code along the way, seemed very low-hanging fruit to me.

Thorium:
> To work properly, either (a) the database would need to acquire full table-locks on every SELECT

...or intelligently know if the transaction contains write operations and acquire the lock based on that.
What is the transaction isolation level you are using?

piaste:
> ...or intelligently know if the transaction contains write operations and acquire the lock based on that. 

Which it cannot guess until the user actually sends a write query. Only the application knows that it might send an INSERT at some point later in the transaction, so it would need to manually acquire the lock (through LOCK TABLE, SELECT FOR UPDATE, or similar commands).

> What is the transaction isolation level you are using?

On Postgres 'read committed', on MSSQL... probably _all of them_ somewhere (that database has hundreds of physical deployments over a decade, by dozens of IT folks). But raising it to `serializable`, temporarily or even permanently, wouldn't help:

 * PostgreSQL uses predicate locking. It would run both SELECTS concurrently, then when one INSERT executes, it would detect that that execution had affected the result of the other transaction, and forcefully rollbacks the latter. In effect you have to deal with a 40001 serialization failure exception instead of a 23505 unique constraint violation exception, not much of an improvement.

 * SQL Server uses range locking, so it actually doesn't guarantee true serializability in this case. Each SELECT would acquire a read-only lock against any row that it *could* have returned (existing or not), but both SELECTs would still be allowed to read concurrently, and both will see no existing row. The locks would then force the second INSERT to wait until the first transaction has completed, but when it does, it will throw a duplicate key exception.

 * SQLite behaves basically like SQL Server, except that the locks are held on the whole database, but since it still has a distinction between read-only (SHARED) and write locks, it would run into the same outcome.

Thorium:
Should I merge this?

piaste:
@Thorium 

Yes, please. I've been using this feature for a while without issues, and it's entirely optional anyway.

The rebasing causes some fake diffs to show up from previous. I've copied the branch into a new PR so you can see only the real diffs. Feel free to merge either one and close the other, as you prefer.

PR #556, opened 2018-06-23
Extend SQLEntity.MapTo to allow mapping when field names are different than column names

francotiveron:
When using SQLEntity.MapTo, the target type (Record or Class) must have field or properties named as the entity columns. Wit this change, a new attribute (MappedColumnAttribute) can be used to decorate them so that the entity column names can be different.


Thorium:
Excellent! May I ask that you add this feature to documentation also?

https://github.com/fsprojects/SQLProvider/blob/3b6821f963952f4897db95f3e3be870c1c6ac329/docs/content/core/mappers.fsx

francotiveron:
Documentation updated

Thorium:
Thanks, merged & released.

PR #558, opened 2018-06-28
#504: SQL2008 compatibility for paging + MSSQL CI

### LINKED ISSUE #504: SQL Server 2008 pagination skip and take
### Description

Provider uses OFFSET FETCH clause for handling SKIP/TAKE, and as stated in code comments "this only works in >=SQL2012"

### Repro steps

1. Table SQL
```
CREATE TABLE [NU-SOFT].[Fixture](
	[Id] [int] IDENTITY(1,1) NOT NULL,
	[Name] [nvarchar](20) NOT NULL)
```
2. Query
```
query {
    for f in ctx.NuSoft.Fixture do
    where (f.Name =% "%a")
    sortBy f.Name
    skip 2
    take 3
}
```

### Expected behavior

Expected SQL for 2008 would be:
```
WITH CTE AS ( 
	SELECT 
		[f].[Id] as 'Id',
		[f].[Name] as 'Name' , 
		ROW_NUMBER() OVER(ORDER BY  [f].[Name]  ) AS RN  
	FROM [NU-SOFT].[Fixture] as [f] 
	WHERE (([f].[Name] LIKE @param1))
)
SELECT 
	[f].[Id] as 'Id',
	[f].[Name] as 'Name' 
FROM CTE [f] 
WHERE RN BETWEEN 3 AND 5
```

### Actual behavior

Actual SQL generated by provider:
```
SELECT 
	[f].[Id] as 'Id',
	[f].[Name] as 'Name' 
FROM [NU-SOFT].[Fixture] as [f] 
WHERE (([f].[Name] LIKE @param1))
ORDER BY [f].[Name] 
OFFSET 2 ROWS FETCH NEXT 3 ROWS ONLY
```

### Related information 

SQL Server 2008

piaste:
Fixes #504 using @devanarch's solution.

Enables SQL Server testing on AppVeyor. Compilation happens against SQL2008R2, then tests are run against both SQL2008R2 and SQL2017.

The CI highlighted a problem with the previous implementation - if the same schema exists on two different servers, it would remember the MSSQL version of whichever one it connected to first. Since this is not an unreasonable scenario - say you're doing an ETL or data sync job between 'old server' and 'new server' - I changed the version cache to a dictionary by connection string.


Thorium:
Thanks!

Issue #552, opened 2018-06-07
Runtime System.Exception when ContextSchemaPath is defined


keyset:
### Description
When the ContextSchemaPath locally cached schema file is defined and present, SQLProvider returns `System.Exception: !` when performing database operations at runtime.

### Repro steps

1. Generate a local schema file with ContextSchemaPath as [described in the documentation](https://fsprojects.github.io/SQLProvider////core/parameters.html#ContextSchemaPath)

2. Debug the project

3. Perform a database operation. The following exception is raised:
```
System.Exception: !
   at FSharp.Data.Sql.Providers.MySql.findClrType@91-1.Invoke(String _arg1)
   at FSharp.Data.Sql.Providers.MySql.createParam[a](String name, Int32 i, a v)
   at <StartupCode$FSharp-Data-SqlProvider>.$Providers.MySql.clo@350-280.Invoke(Tuple`2 tupledArg)
   at Microsoft.FSharp.Collections.SeqModule.Fold[T,TState](FSharpFunc`2 folder, TState state, IEnumerable`1 source)
   at FSharp.Data.Sql.Providers.MySqlProvider.createInsertCommand(IDbConnection con, StringBuilder sb, SqlEntity entity)
   at <StartupCode$FSharp-Data-SqlProvider>.$Providers.MySql.FSharp-Data-Sql-Common-ISqlProvider-ProcessUpdates@949-3.Invoke(SqlEntity e)
   at Microsoft.FSharp.Collections.SeqModule.Iterate[T](FSharpFunc`2 action, IEnumerable`1 source)
   at FSharp.Data.Sql.Providers.MySqlProvider.FSharp-Data-Sql-Common-ISqlProvider-ProcessUpdates(IDbConnection con, ConcurrentDictionary`2 entities, TransactionOptions transactionOptions, FSharpOption`1 timeout)
   at <StartupCode$FSharp-Data-SqlProvider>.$SqlRuntime.DataContext.f@1-69(SqlDataContext __, IDbConnection con, Unit unitVar0)
   at FSharp.Data.Sql.Runtime.SqlDataContext.FSharp-Data-Sql-Common-ISqlDataContext-SubmitPendingChanges()
   at MyProject.DoDatabaseStuff(DbRow _row) in C:\Users\Me\Code\MyProject\Db.fs:line 65
Stopped due to error
```

### Expected behavior
[...]

PR #560, opened 2018-07-09
Fix context schema path + cleanup GetDataContext() TP

piaste:
Pre-submitting for review / feedback. Tests are green but I should add a few more to ensure that the designtime - runtime differences do not hold any surprises handled; the current CI tests do change the connection string and occasionally the timeout, but nothing else. And the issue with schema files on dev machines is going to need manual testing.

The urgent concern is addressing #552 . This should be solved by passing `""` as the `contextSchemaPath` in the runtime quotation [here](https://github.com/fsprojects/SQLProvider/compare/fsprojects:4e1943a...piaste:159db96#diff-ce03f5aad6f04b98307bc819c893a50fR797), preventing runtime loading. 

The rest is a cleanup of that function, which @Thorium pointed out was in sore need of it. This version currently takes the lazy approach and just generates all 32 possible overloads of `.GetDataContext()` (which should ensure no breaking changes). But if one wanted we could manually pick and choose the available overloads, by replacing the five for-loops [here](https://github.com/fsprojects/SQLProvider/compare/fsprojects:4e1943a...piaste:159db96#diff-ce03f5aad6f04b98307bc819c893a50fR768) with an explicit list:

```fsharp

let overloads = [|
  [| UserProvided customConnStr; Default defaultResPath; Default defaultTransOpts; Default defaultCmdTimeout; Default defaultSelectOps |]
  [| UserProvided customConnStr; UserProvided customResPath; Default defaultTransOpts; Default defaultCmdTimeout; Default defaultSelectOps |]                        
  // etc.                                                
|]                        

for actualParams in overloads do 
 // continue as below
```


Thorium:
The current problem is that the caches don't span over processes: Visual tudio has its own and each compilation of fsc.exe will have empty cache. So with a semi-large project having a remote db, build takes minutes, not seconds.

Thorium:
This is marked as WIP. 1) I like the idea of parameter combinations. Is it baackward compatible? 2) What is still missing, can I already merge this?

piaste:
>  I like the idea of parameter combinations. Is it baackward compatible? 

In testing I found that the full combinatorial generation was a breaking change in a bad way, because it led to ambiguous overloads (like `(connString : string)` and `(resPath : string)`) which mandated the use of named parameters everywhere.

I added some sugar instead, so now the overload list is explicit but [very maintainable](https://github.com/fsprojects/SQLProvider/pull/560/files#diff-ce03f5aad6f04b98307bc819c893a50fR774).

> What is still missing, can I already merge this?

I don't have any more code changes planned, but I would still like to do a couple of manual tests, I hope I can get a chance to do them before the week's end. In particular, this is the procedure I have in mind:

1) Save a context schema .json file
2) Close VS, drop a column in the database that was used by the application
3) Restart VS, compile the application, should use the now-outdated schema and compile
4) Add the column back to the database
5) Run the application, should not throw
6) Manually put garbage in the schema .json file, should still not throw
7) Delete the .json file, should still not throw

piaste:
Ok, I've rebased on master and done the manual testing I described above, I'm OK with merging this.

Tarmil:
Hi, is there an ETA for when this will drop on NuGet? I would love to use ContextSchemaPath but I currently can't because of this issue.

Thorium:
Sorry, I checked the 3 changed files and thought that this won't actually change anything, maybe I missed something. I'll release now, 1.1.50

Tarmil:
The issue is somewhat subtle -- I'm compiling for .NET Core, so I need to use the .NET Framework compiler, and the serialization format is apparently different between the two for some reason, so the runtime failed to parse the file created at compilation.

It works fine now with 1.1.50, thanks!

Issue #312, opened 2016-08-09
How to execute stored procedure with OUTPUT parameter?


StefanBelo:
I have got MSSQL Express stored procedure with OUTPUT parameter, but the invoke method maps this parameter like entry one, not with reference.

Does SQLProvider support such operation?

pezipink:
Hello, yes we do support OUT parameters in a round-about manner.  I believe they are returned as part of the result set(s) of the sproc rather than attempting to use byref params which are kind of nasty in F#.  

//cc @colinbull

StefanBelo:
Thanks, but it seems I am missing something. When executing stored procedure by calling Invoke, the Invoke method returned value is unit. So what you suggest is that I must changed stored procedure to return one record containing my OUTPUT value?

pezipink:
No, it should work for you.  It may be you have ran into an edge-case, I will try and have a look at it a bit later for you unless someone else gets there first.  Is it possible to share the definition of the procedure?  or the salient bits of it at least.

StefanBelo:
Here is my database schema, including all tables, views and stored procedures:

[http://bfexplorer.net/Content/Downloads/Scripts/MyRaceDatabaseScript.zip](http://bfexplorer.net/Content/Downloads/Scripts/MyRaceDatabaseScript.zip)

StefanBelo:
Any news? I tried SqlProgrammabilityProvider and it detects OUTPUT parameters for all my stored procedures.

pezipink:
I have examined this and it looks like we are simply not supporting output parameters properly.  This will need a bit of work since the sproc code is fairly complex - I'll see if I can get a prototype going later in the day, but I might not be able to.

@colinbull I am not missing something silly am I?

StefanBelo:
Thank you, I will check your progress later, for now will use SqlProgrammabilityProvider.

colinbull:
@pezipink I know other providers do support out parameters, but looking at  https://github.com/fsprojects/SQLProvider/blob/master/src/SQLProvider/Providers.MsSqlServer.fs#L240, it would appears that we explicitly exclude them. I cannot remember for the life of me why this is, but I remember having (what seemed like) a good reason for it at the time.

Issue #171, opened 2015-11-20
Sproc and output parameters


xenocons:
Hello,

it appears that .Invoke does not properly utilise output parameters. Within Providers.MsSqlServer.fs,  when executeSprocCommand is called, 'returnCols' appears to be null, so output parameters do not get populated.

firstly, here is the stored proc:
```
ALTER PROCEDURE [dbo].[XXYY]
	@hash NVARCHAR(MAX),
	@exists bit OUTPUT
	
AS BEGIN
	IF EXISTS (SELECT Hash FROM [Storage].ABC
		WHERE Hash = @hash)
	BEGIN
		SELECT @exists = 1
	END
	ELSE 
		SELECT @exists = 0
END
```

```
let mutable out = false
db.Procedures.XXYY.Invoke ("test",out)
```
Result:
```
exec x.dbo.XXYY @hash=N'test'
go
```

```
> System.Data.SqlClient.SqlException (0x80131904): Procedure or function 'XXYY' expects parameter '@exists', which was not supplied.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString)
   at System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, SqlDataReader ds, Boolean describeParameterEncryptionRequest)
[...]

PR #564, opened 2018-07-21
Fix for mssql sprocs with output parameters

OlofBlomqvist:
Not sure if this is good enough, but based on comments in issue #312 and #171, I think this is how output variables are expected to be returned.

sqlprovider_sprocout2

sqlprovider_sprocout1

image

Also, this is my first PR - sorry if im doing it wrong!


Thorium:
Thanks!

PR #566, opened 2018-08-11
Use sproc params with context schema cache

Thorium:
Could it be this simple...?

### DIFF
--- a/src/SQLProvider/SqlDesignTime.fs
+++ b/src/SQLProvider/SqlDesignTime.fs
@@ -97,12 +97,16 @@ type SqlTypeProvider(config: TypeProviderConfig) as this =
             lazy
                 match con with
                 | Some con -> prov.GetSprocs con
-                | None -> []
+                | None -> prov.GetSchemaCache().Sprocs |> Seq.toList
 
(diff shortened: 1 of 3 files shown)


object:
Hmm, looks promising :-)

Thorium:
Turns out it wasn't, but it's a good start anyway...

Thorium:
Now it should be working!

Thorium:
Working.. But as commented to the open issue about ContextSchemaPath: Another problem I found out was that when I have a two-project-solution, only the active project's schema is saved when I call the save method.

I could deal with this by saving 2 files separately and merging them somehow. How should this work? E.g. Save as is, but when loading, having a semicolon-separator in the context-schema-path static parameter?

object:
Ideally it should save the schema across the projects, but I must admit I didn't test the scenario with multiple projects referencing each own part of the schema. I deliberately wanted to release the simple scenario first to collect feedback, because the whole concept is quite unexplored and I still don't have strong opinion on how the workflow should look.

Thorium:
For now, I created one public method to solve my problem: 
```fsharp
FSharp.Data.SqlCommon.OfflineTools.mergeCacheFiles("c:\\project1.json", "c:\\project2.json", "c:\\both.json")
```

That will do.

PR #567, opened 2018-08-12
[MSSQL] Add brackets around the OUTPUT column name of the INSERT command to …

DieselMeister:
…avoid error on columns with reserved names.

For example a SQL-Table has a id column named "key" (I think it was the backing table of the identity server 3). The insert command produce an error, because "key" is a reserved key word in SQL.

To avoid that, using brackets around the colmun name.


Thorium:
Thanks!

PR #603, opened 2019-03-01
Initial nested exists sub-query support.

Thorium:
```fsharp
    let qry = 
        query {
            for o in dc.Main.Orders do
            where(query {
                    for od in dc.Main.OrderDetails do
                    exists(od.OrderId = o.OrderId)
                })
            select o.OrderId
        } |> Seq.toList
```
...produces:

```sql
SELECT [o].[OrderID] as 'OrderID' 
FROM main.Orders as [o] WHERE (
   (EXISTS (
        SELECT [od].[Discount] as 'Discount',[od].[OrderID] as 'OrderID',[od].[ProductID] as 'ProductID',[od].[Quantity] as 'Quantity',[od].[UnitPrice] as 'UnitPrice' 
        FROM main.OrderDetails as [od] 
        WHERE (([od].[OrderID] = [o].[OrderID]))
   ))
) 
```


Thorium:
The whole baseAlias when "" is full-tablename is messy and should be replaced with normal alias, e.g.
`for o in orders` should have alias of `o` and not `Orders`.


This is an issue when there are simple sub-queries that have no aliases but some aliasses can be inherited from the main-query. (free variables)

Thorium:
The recent alias resolve fix did fix also the issues with this branch.

Issue #575, opened 2018-09-12
.NET Core/.NET Standard support (Tracking issue)


dsyme:
I'm creating a tracking issue to help assess the status of .NET Core/.NET Standard support for SQLprovider.  If people can help me define what we need to assess that would be great.

Below I've partially listed some of the scenarios.  The matrix is unforuntately large because of the variety of compile-time tools that may end up hosting the type provider TPDTC, but hey.

If there is a `?` it means I don't know what is the highest priority thing to assess for that scenario.

What I'm really looking for is help to know what 1 or 2 scenarios we should aim to "complete" next

| Description | Compile-time Tool        |  Runtime target           | Databases  | OS | Works? | Docs? | CI? |
|:-------------|:-------------|:-------------|:-------------:|:-----:|:-----:|:-----:|:-----:|
| Target .NET Framework 4.x using VS2017     | devenv.exe | net461 | postgres, mssql, ...  | Windows  |  👍  |  👍  |  👍 |
|                                 | fsc.exe (desktop) |  |   |   | 👍  |  👍 |  👍 |
|                                 | fsc.exe (dotnet) |  |  ? |  Windows |   |   |   |
| Target .NET Core App 2.0 using VS2017     | devenv.exe | netcoreapp2.0  | ?  | Windows  |   |   |   |
|                                 | fsc.exe (desktop) | |  ? |  |   |   |   |
|                                 | fsc.exe (dotnet) |   |  ? |  |   |   |   |
| Target .NET Standard 2.0 using VS2017     | devenv.exe | netstandard2.0 |  ?  |  |   |   |   |
|                                 | fsc.exe (desktop) | |  ? |   |   |   |   |
|                                 | fsc.exe (dotnet) |  |  ? |   |   |   |   |
| Target .NET Core App 2.0 using .NET SDK | fsc.exe (dotnet) | netcoreapp2.0  | ?  | Linux |   | |   |
|  |  |  |   | OSX|   | |   |
| Target Mono using Mono | fsc.exe (desktop) | net461 | postgres  | Linux | 👍   | 👍  |  👍 |
|                                           |             |   |    | OSX | 👍   | 👍  |   |

Thorium:
.NET Framework supported databases:
- Microsoft SQL Server, Postgres, MySQL / MariaDB, SQLite, FireBird, Oracle, Odbc, Access

.NET Core:
- In theory all could be supported, but I haven't seen working data drivers for Oracle, Odbc and Access. 
[...]

PR #611, opened 2019-04-23
[HACK] Replace references to `None` basetype to `Some typeof<obj>`

rudihorn:
This is a PR to solve the issue I mention in #575. Somehow my editor is configured to automatically remove trailing spaces so it has done this automatically upon saving the file as well.

Essentially this PR replaces all values of `None` for the basetype value of `ProvidedTypeDefinition` to `Some typeof<obj>`. To me this seems like a hack as there seems to be something fundamentally wrong with type resolution with .NET Core and F#, as there seems to be countless projects with similar issues revolving library resolution, but this is a hack which somewhat fixes it for SQLProviders for the moment.

It would be nice if this could be pulled soonish, as the fsharp language server doesn't seem to like loading SQLProviders from a locally saved project but I think it handles the nugget version reasonably.

### DIFF
--- a/src/SQLProvider/SqlDesignTime.fs
+++ b/src/SQLProvider/SqlDesignTime.fs
@@ -13,15 +13,15 @@ open FSharp.Data.Sql
 open ProviderImplementation.ProvidedTypes
 
 type internal SqlRuntimeInfo (config : TypeProviderConfig) =
-    let runtimeAssembly = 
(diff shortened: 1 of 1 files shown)


Thorium:
Thanks!

Thorium:
Released in NuGet version 1.1.62

rudihorn:
Thanks!

PR #618, opened 2019-06-02
Handle "Views" in SQLite schema

iNecas:
Without this patch, I was getting the following error:

```
error FS3033: The type provider 'FSharp.Data.Sql.SqlTypeProvider'
reported an error: Not supported. This custom getSchema will be removed
when the corresponding System.Data.Common interface is supported by
the connection driver.
```

It turned out the `customGetSchema` method was not counting on `"Views"`
in `name` variable. This patch should support this by splitting
current handler for `"Tables"`, that seemed to handle both tables and
views in the past.

This patch also enhances the error message, so that next time, it would
include the name of the unsupported field right in the message, like
this:

```
error FS3033: The type provider 'FSharp.Data.Sql.SqlTypeProvider'
reported an error: Not supported [ Views ].
# ...
```

### DIFF
--- a/src/SQLProvider/Providers.SQLite.fs
+++ b/src/SQLProvider/Providers.SQLite.fs
@@ -20,14 +20,25 @@ type internal SQLiteProvider(resolutionPath, contextSchemaPath, referencedAssemb
     /// This is custom getSchema operation for data libraries that doesn't support System.Data.Common GetSchema interface.
     let customGetSchema name conn =
         let dt = new DataTable(name)
+        let updateDataTableByTableOrView (masterType:string) =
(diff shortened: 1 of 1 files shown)


iNecas:
An example app reproducing the issue, as well as proving the patch fixing it, can be found here https://github.com/iNecas/fsharp-sqlite

iNecas:
SRY for whitespace changes, I'm ok with removing those if necessary (although would prefer cleaning up the training spaces)

Thorium:
Thanks!

Thorium:
Released in Nuget package 1.1.64

iNecas:
Thanks @Thorium!

PR #649, opened 2019-11-08
Fix "timestamp with timezone" mapping into DateTimeOffset

Zaid-Ajaj:
Columns of type "timestamp with timezone" should have time DateTimeOffset instead of `DateTime` because otherwise you lose the timezone information


Zaid-Ajaj:
Closing because it seems I don't understand timezones quite yet, thanks @kerams for explaining it to me once again in [this comment](https://github.com/demetrixbio/FSharp.Data.Npgsql/pull/49#issuecomment-552107609)

PR #674, opened 2020-04-27
ctx.``Design Time Commands``.ClearDatabaseSchemaCache(), without restarting your IDE

Thorium:
Database schema is cached lazily meanwhile you use it. Sometimes your schema
may change meanwhile you develop your code, and you would like to invalidate
the database schema cache without restarting your IDE. This can be done via method 
`.``Design Time Commands``.ClearDatabaseSchemaCache().`


granicz:
In general, most users will come with the expectation/idea that there is something they can do to the *TP invocation* to trigger the TP to invalidate any cached info it gathered. I would suggest adding an argument, such as `Invalidate=true/false` as well.

(There is a typo in the error message in your commit.)

Thorium:
@granicz I'm not familiar with this mechanism, I tried to add new argument `ProvidedStaticParameter("Invalidate",typeof<bool>,false)` and then handling:

```fsharp
        let invalidateCache = args.[12] :?> bool
        if invalidateCache then
            (this :> ITypeProvider).Invalidate.Add(fun _ ->
                DesignTimeCache.cache.Clear()
            )
```

to here: https://github.com/fsprojects/SQLProvider/blob/a32ca824a4d36d665b3a346456a6df9575bc15da/src/SQLProvider/SqlDesignTime.fs#L945

...but it didn't work.

PR #677, opened 2020-05-12
Build update (FAKE and document generation fix)

Thorium:
- Fixed build where document generation failed on some environments.
- Moved Paket to use separate Build group dependencies 
- Updated FShap.Formatting and FAKE and some other dependencies


Thorium:
Status: Locally has been working fine, but the CI fails for some weird reasons:
AppVeyor: Old Test solution fails to build
Travis: Cannot open assembly 'packages/build/FAKE/tools/FAKE.exe': No such file or directory.

Thorium:
This is out-dated and implemented already other ways.

PR #681, opened 2020-06-05
Change issuestats to shields badge

seonggwonyoon:
## Proposed Changes

Badges were broken in README.md. The issuestats.com used previously did not work properly because the domain was sold, and the badge was changed to Shields to work properly.


Thorium:
Thanks!

Issue #692, opened 2020-10-01
SQLProvider does not support PublishSingleFile switch in .NET 5


pascallapradebrite4:
**Describe the bug**
Executables produced with .NET 5 SDK RC 1 crash when built with the `PublishSingleFile` switch. The stacktrace produced is the following:

```
Unhandled exception. System.Exception: Unable to resolve assemblies. One of MySql.Data.dll, MySqlConnector.dll (e.g. from Nuget package MySql.Data) must exist in the paths: 
 /home/vsts/work/1/s/Database
/home/vsts/.nuget/packages/mysqlconnector/1.0.0/lib/net5.0
/home/pascallaprade 
Details: 
resolutionPath directory doesn't exist:/home/vsts/work/1/s/Database
   at Microsoft.FSharp.Core.PrintfModule.PrintFormatToStringThenFail@1433.Invoke(String message)
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
--- End of stack trace from previous location ---
   at System.Lazy`1.CreateValue()
   at FSharp.Data.Sql.Providers.MySql.createConnection[a](a connectionString)
   at FSharp.Data.Sql.Providers.MySqlProvider.FSharp-Data-Sql-Common-ISqlProvider-CreateConnection(String connectionString)
   at <StartupCode$FSharp-Data-SqlProvider>.$SqlRuntime.DataContext.addCache@37-1.Invoke(Unit unitVar)
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.get_Value()
   at FSharp.Data.Sql.Runtime.SqlDataContext..ctor(String typeName, String connectionString, DatabaseProviderTypes providerType, String resolutionPath, String[] referencedAssemblies, String runtimeAssembly, String owner, CaseSensitivityChange caseSensitivity, String tableNames, String contextSchemaPath, OdbcQuoteCharacter odbcquote, SQLiteLibrary sqliteLibrary, TransactionOptions transactionOptions, FSharpOption`1 commandTimeout, SelectOperations sqlOperationsInSelect)
   at MachineComputerManager.Database.Context.getContext(String connectionString)
   at MachineComputerManager.Database.Database.getContext(ConfigType config)
[...]

PR #694, opened 2020-10-12
Added capability to load drivers from AppDomain

pascallaprade:
## Proposed Changes

Proposed fix for issue #692.

Because .NET 5's PublishSingleFile loads assemblies in memory instead of from files, SQLProvider needs to be able to load the drivers from the memory as well.

The current driver resolution logic remains untouched, but if it fails to load the expected assemblies, we then try to load them from the AppDomain instead.

This requires the users of SQLProvider to ensure that their driver is loaded into the AppDomain before creating the context. I tried other ways, such as getting the referenced assemblies of the executing assembly, but it would not work with a solution where the executing assembly relies on another project for data access through SQLProvider, so simply using the AppDomain seemed like the simplest way to support both single-project and multiple-projects solutions.

## Types of changes

## Checklist

Note: I would be happy to provide tests as well, but I am uncertain as to how to proceed. If you can push me in the right direction for that, I'll add them and update the pull request.

## Further comments

As mentioned in the proposed changes, the current changes require users to statically "use" the driver in their own code for it to be loaded in the AppDomain. This creates an undesirable line of code to be added in the user's project, which I found to be acceptable, but instead of that, it could be imaginable that the assembly is passed by the user into the GetContext method call instead. It would be a bit cleaner than adding a typeof call I think, but I am unsure how I would make the change. If you prefer this second approach, I'll need a bit of help to get me in the right direction as to where changes would need to be made.

### DIFF
--- a/docs/content/core/parameters.fsx
+++ b/docs/content/core/parameters.fsx
@@ -3,8 +3,27 @@
 (*** hide ***)
 #I "../../../bin/net451"
 (*** hide ***)
+#r "../../../packages/scripts/MySqlConnector/lib/net45/MySqlConnector.dll"
+(*** hide ***)
 #r "FSharp.Data.SqlProvider.dll"
 open FSharp.Data.Sql
+
+(*** hide ***)
+[<Literal>]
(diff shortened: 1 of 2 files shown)


Thorium:
Thanks!

Issue #698, opened 2020-12-10
SSDT options


JordanMarr:
I am considering creating an SSDT provider that would allow people using SSDT to get their table, column and relationships (and possibly views as well) from SSDT scripts. I would imagine this to be something like Providers.MsSqlServer.SSDT.fs so that it could be based off the the MsSqlServer provider but it would parse the SSDT scripts instead of doing a schema query.

I just wanted to see if there was any interest in having that as an option for SQLProvider before I kick the tires any more.

Thorium:
Sounds good, I'll merge and release happily if you get it done.

JordanMarr:
I'm getting an error when I run the build script: 

**"C:\_github\SQLProvider\src\SQLProvider\SqlProvider.fsproj : error MSB4057: The target "Rebuild" does not exist in the project."**



Here's the full output:

Checking Paket version (downloading latest stable)...
Paket.exe 5.257.0 is up to date.
Paket version 5.257.0
The last restore is still up to date. Nothing left to do.
Performance:
 - Runtime: 110 milliseconds
This runner is now obsolete with FAKE 5, please upgrade to the new .Net Core runner. See https://fake.build/fake-migrate-to-fake-5.html
To remove this warning you can append the '--removeLegacyFakeWarning' argument or set the 'FAKE_NO_LEGACY_WARNING' environment variable to 'true'
Building project with version: LocalBuild
Shortened DependencyGraph for Target All:
<== All
   <== GenerateHelp
      <== GenerateReferenceDocs
         <== CleanDocs
            <== RunTests
               <== BuildCore
                  <== Build
                     <== AssemblyInfo
                        <== Clean

The running order is:
  - Clean
  - AssemblyInfo
  - Build
  - BuildCore
  - RunTests
  - CleanDocs
  - GenerateReferenceDocs
  - GenerateHelp
  - All
Running build with 1 worker
Starting Target: Clean
C:\_github\SQLProvider\bin does not exist.
C:\_github\SQLProvider\temp does not exist.
Finished Target: Clean
Starting Target: AssemblyInfo (==> Clean)
Created AssemblyInfo file "src/SQLProvider/AssemblyInfo.fs".
Finished Target: AssemblyInfo
Starting Target: Build (==> AssemblyInfo)
Building project: C:\_github\SQLProvider\SQLProvider.sln
[...]

PR #700, opened 2020-12-19
SSDT Provider

JordanMarr:
## Proposed Changes

SSDT provider as discussed in #698 

## Types of changes

## Checklist

## Further comments
If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

I haven't added my unit tests to the sln because I had a hard time getting the build and existing tests to compile and run. Hopefully the tests pass -- fingers crossed! 😁
I did add a new documentation page, and I *think* I linked to it on the main page, but I wasn't able to build the docs to verify.


Thorium:
Huge work, thanks! It seems Appveyor CI was working before but is not with this build, however I think we can try to fix it later...

Thorium:
This is now in NuGet v. 1.1.95
Let me know how it's working...

JordanMarr:
> This is now in NuGet v. 1.1.95
> Let me know how it's working...

It's working!
I dropped it into my existing project that used MSSQLSERVER provider and that worked.

Then I swapped it out with the new MSSQLSERVER_SSDT. 
The tricky part though is that I'm using VS2019 with NuGet PackageReferences, so when I added the `Microsoft.SqlServer.Management.SqlParser.dll`, I could reference it in "packages" because it stores it in a global cache:
"C:\Users\jmarr\.nuget\packages\microsoft.sqlserver.management.sqlparser\160.20216.14"
I found the path by clicking on my project Dependencies/Packages/ and then looking at the package properties. 
So then I copied the "netstandard2.0" version of the dll into my project folder, added that dll to my .gitignore file, and then finally was able to reference my project folder as the ResolutionPath.  
After that, it works like a charm!

So I'm thinking that maybe I need to update the docs page with a little more guidance on how to set that up properly.
If you get the wrong ResolutionPath, or worse, if you try to reference the net462 version of the dll and you're using a netstandard project, it will give all kinds of crazy error messages when it tries to parse at design time.  

But the good news is that it does work when setup properly!

PR #701, opened 2020-12-20
Listed SSDT page on index page, ...

JordanMarr:
...and added detailed instructions for properly setting the SSDT ResolutionPath.

## Proposed Changes
I just wanted to add the SSDT provider to the index help page, and more importantly, I added some detailed instructions for setting up the ResolutionPath (since that is so crucial to getting it to work). 

## Types of changes

I only updated the docs.

### DIFF
--- a/docs/content/core/mssqlssdt.fsx
+++ b/docs/content/core/mssqlssdt.fsx
@@ -12,6 +12,8 @@ open System
 
 # MSSQL SSDT Provider
 
+The SSDT provider allows types to be provided via SQL Server schema scripts in an SSDT project. No live database connection is required!
+
 ## Parameters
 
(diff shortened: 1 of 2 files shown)


Thorium:
Thanks!

Thorium:
Docs updated... I think the stored proc support will be quite high priority.
Maybe you could introduce this as an entry to fsharp advent calendar, if not too late?

JordanMarr:
> Docs updated... I think the stored proc support will be quite high priority.
> Maybe you could introduce this as an entry to fsharp advent calendar, if not too late?

Good point on the stored procs...
I don't think they'll be too bad, so I can probably knock them out today.

PR #703, opened 2020-12-25
SSDT script discovery, views and sprocs

JordanMarr:
## Proposed Changes

* Improved SSDT .sql script discovery: `SsdtPath` can now point to a `.sqlproj` project file and will get paths to all .sql scripts marked as Build.
* `SsdtPath` can still be set to a root folder instead, but now it is not reliant on subfolder conventions; instead, it will recursively search for .sql files under the given `SsdtPath` and analyze them to determine script type.
* Added support (or at least basic support) for stored procs
* Fixed a bug in the view parser where aliased columns were using the referenced table column name
* If a .sql file fails parsing for any reason, the file name will be displayed to user when they hover over the type provider to help with troubleshooting.
* Updated SSDT docs

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

* Adding .sqlproj support for SsdtPath is a non-breaking new feature. 
* Changing SsdtPath behavior when set to a folder path to recursively scan for .sql files and auto-analyzing their types (table, view, sproc) is potentially breaking change since previously it didn't analyze scripts and expected convention based folders. 
I think it's a minor risk though.
* New feature is implementation of sprocs

## Checklist

## Further comments

I don't think I have added my SSDT test project to the test sln yet for the build script to execute. I'll try to remember to do that next time.


Thorium:
Thanks! Released in 1.1.97

PR #707, opened 2021-01-10
Ssdt annotations

JordanMarr:
## Proposed Changes

* Added new Type Annotation feature that allows annotating types for view columns via in-line comments (added as a work-around since the .dacpac can't resolve computed columns).
* Added column descriptions that indicate if a view column has a type annotation (or if it needs one).
* Fixed confusing error message when the SsdtPath was invalid (which said it was looking for it in the .nuget global cache folder)
* Updated SSDT docs with new Type Annotation features
* Updated SSDT docs with more information on setting the SsdtPath for design-time and for deployment.
* Added support for table computed columns so that they now exist with a type of `obj`. ~~(Unfortunately, it looks like the .dacpac doesn't let me query comments on tables the same way it does for views, so I'm not yet sure how to do inline comment type annotations for computed table columns yet.)~~
* Added Type Annotations for computed table columns.

## Types of changes

## Checklist


JordanMarr:
Actually it looks like the .dacpac will include the comment for a computed table column, but only if the comment is contained within the parenthesis (which can easily be documented). So hold off on merging this until I can implement annotations for computed table columns.

JordanMarr:
I think this PR is now ready! I'll fix the stored procs issue later.

Thorium:
Thanks!
Released in 1.1.99

Issue #645, opened 2019-10-16
TypeProvider not working with dotnet core 3.0


tforkmann:
### Description

I'm trying to use the typeprovider with netcoreapp3.0 (3.0.100). But it is not working for me.
I alread tried to reference System.Data.SqlClient.
Ionide is failing as well.

### Repro steps

1. Set up Dotnet core app 3.0

2. Add SqlProvider

3. Try to get datacontext

```fs
open FSharp.Data.Sql
open System
open Config

/// Query DME
let [<Literal>]  CompileTimeConnectionString = "myconnectionString"

type SqlDme = SqlDataProvider<Common.DatabaseProviderTypes.MSSQLSERVER, ConnectionString=CompileTimeConnectionString, UseOptionTypes=true> //, ContextSchemaPath = ContextSchemaPath>
````

### Expected behavior

TypeProvider should just work

### Actual behavior

I get following error message when building the project:
Could not load file or assembly 'System.Data.SqlClient, Version=4.6.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Das System kann die angegebene Datei nicht finden. [C:\Users\tforkmann\Documents\1_Tests\PreveroFeed\src\Server\Server.fsproj]
![image](https://user-images.githubusercontent.com/4168951/66926959-3ed84200-f02f-11e9-83cc-d2f42af5730f.png)

### Known workarounds

1. Add FSharp.Compiler.Tools
2. Add to fsproj file
```<DotnetFscCompilerPath></DotnetFscCompilerPath>```

### Related information 

* MS Sql
* Windows 10
* netcoreapp3.0

tforkmann:
#580

forki:
Id there a general strategy for type providers in dotnet sdk 3.x? Since we don't get any new fsharp.compiler.tools packages anymore - what's the replacement?

/cc @dsyme @cartermp

Bjorn-Strom:
Having this same problem on mac.

cartermp:
@forki There is no replacement, as that was never a supported scenario in the first place. I'd encourage never installing that package.

forki:
Reality is that this was the only way to get things working. But it's not really important anymore. The question is how can we proceed?

cartermp:
Realistically there's probably a bug in either the type provider, the provider sdk, or F# compiler that needs to be resolved. Using an unsupported package containing a very old .NET Framework-only version of the F# compiler isn't a scenario you should be considering here.

[...]

PR #708, opened 2021-01-12
Fixes error for netstandard where it tries to load SqlClient 4.6.1.0

JordanMarr:
## Proposed Changes

In reference to issue #645 where a .NET Core project fails with an error message about not being able to load System.Data.SqlClient Version=4.6.1.0 - and also my similar issue where the SSDT Provider gives the same error from VS Code (but not in VS 2019): this change to the build script fixes the issue for me in VS Code.

## Types of changes

## Checklist

## Further comments

This just copies the netstandard2.0 version of System.Data.SqlClient instead of net461 to the bin/netstandard2.0 output folder.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -255,7 +255,7 @@ Target "CopyFiles" (fun _ ->
     //     copyDotnetLibraries dotnetSdk212
     copyDotnetLibraries "packages/standard/NETStandard.Library/build/netstandard2.0/ref/"
 
-    CopyFile "bin/netstandard2.0" "packages/System.Data.SqlClient/lib/net461/System.Data.SqlClient.dll" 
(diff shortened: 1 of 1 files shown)


Thorium:
A few questions, please:
- Will this break Visual Studio 2017/2019 which are running fsx on .NET framework rather than dotnet?
- Also there has been some issues with System.Data.SqlClient on .Net Core 3.1, that's why they started to do Microsoft.Data.SqlClient.dll, and I think the package versions SQLProvider refernces, System.Data.SqlClient works on .NET and Microsoft.Data.SqlClient works on .Net Core. I've not followed the recent changes on those projects. See https://github.com/fsprojects/SQLProvider/issues/645#issuecomment-639163535

thijstijsma:
Would it make sense to copy it to both places as a workaround?

Thorium:
That's not the target but the source...

PR #709, opened 2021-01-21
SSDT provider support dot (".") in table names, like [dbo].[My.Table]

Thorium:
SSDT provider support dot (".") in table names, like [dbo].[My.Table]


JordanMarr (changes requested):
There is one other place that splits table names: `splitAndRemoveBrackets` on line 15 which is used by `RefTable`. (`RefTable` is used by views to reference back to tables to resolve view columns).

For some reason I made `RefTable` have property members that get `Schema` and `Name` on demand which is inconsistent. It would be better if `Schema` and `Name` were added as record fields and then they could be parsed using your new `splitDottedFullName` function at parse time.

Thorium:
I just wanted to get my schema working, I didn't actually use the table, and this PR was enough. I'll merge this and we can continue improving from here.

Issue #706, opened 2021-01-04
Dacpac location + some testing


Thorium:
Ping @JordanMarr 

I'm testing the dacpac system.
Initially I'll setup the dacpac path as follows:
```fsharp
SsdtPath = __SOURCE_DIRECTORY__ + "database/bin/Debug/database.dacpac",
```
...the type provider says `error: SsdtPath must point to a dacpac file`. Which I thought it does, but obviously something is wrong, so it would be nice to display the path it tries when fails.

The second try, absolute path, this is not as nice, as this won't work in both Mac and Windows:

```fsharp
SsdtPath = @"C:\git\MyProject\database\bin\Debug\database.dacpac",
```

Now the intellisense works great. However when I build my (multi-project) solution, it fails:

```
1>C:\git\MyProject\Program.fs(23,5): error FS3033: The type provider 'FSharp.Data.Sql.SqlTypeProvider' reported an error: Unable to find .dacpac file at: 'C:\git\MyProject\packages\SQLProvider\lib\net451\database.dacpac'
```

Why is it trying to load it from that path?

---

_Edit:_
I copied my dacpac file under my packages path so that the build will find it.
Then I had to restart my VS just that the tables are found in other source files also.

The few things I note not to work:
- View column with `select col1 + col2 as Sum` has type `obj` when it should be `Option<decimal>`
- View column with `select case when 1=0 then datecol1 else datecol2 end as mydate` has type `obj` when it should be `Option<DateTime>`
- Computed columns are not found
- Invoking a stored procedure seems to return `Unit` when it should return `SprocResult` having a ResultSet of `SqlEntity[]`

Other than that looks very good!

JordanMarr:
> ```fsharp
> SsdtPath = _SOURCE_DIRECTORY_ + "database/bin/Debug/database.dacpac",
> ```
> 
> ...the type provider says `error: SsdtPath must point to a dacpac file`. Which I thought it does, but obviously something is wrong, so it would be nice to display the path it tries when fails.
> 

It definitely needs a forward slash directly after __SOURCE_DIRECTORY__:
`__SOURCE_DIRECTORY__ + "/database/bin/Debug/database.dacpac"`

The only way it throws the "SsdtPath must point to a dacpac file" is if `SsdtPath.EndsWith(".dacpac")` fails, which is odd because yours does.  
[...]

PR #711, opened 2021-01-24
Ssdt bugfixes

JordanMarr:
## Proposed Changes

- Creates a more comprehensive `splitFullName` function as a regex (with added unit tests)
- First attempt to fix SSDT stored procs so that they always return a `SprocResult` instead of returning `unit` per request in #706.

## Types of changes

## Checklist

## Further comments

Please let me know if the stored proc change works for your use case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -245,3 +245,4 @@ docs/content/core/northwindEF.db
 .ionide/
 *.svclog
 TestResult.xml
+/tests/SqlProvider.Core.Tests/MsSqlSsdt/MsSqlSsdt.Tests/AdventureWorks_SSDT/AdventureWorks_SSDT.jfm
--- a/src/SQLProvider/Providers.MsSqlServer.Ssdt.fs
+++ b/src/SQLProvider/Providers.MsSqlServer.Ssdt.fs
@@ -1,3 +1,4 @@
+// fsharplint:disable Hints CanBeReplacedWithComposition
 namespace FSharp.Data.Sql.Providers
 
 open System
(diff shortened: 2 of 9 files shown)


Thorium:
For some reason after this PR the test project build fails as:
"System.Exception: NUnit: cannot run tests (the assembly list is empty)."

JordanMarr:
> For some reason after this PR the test project build fails as:
> "System.Exception: NUnit: cannot run tests (the assembly list is empty)."

Is that coming from the MsSqlSsdt.Tests project?

JordanMarr:
> For some reason after this PR the test project build fails as:
> "System.Exception: NUnit: cannot run tests (the assembly list is empty)."

I did make a few small changes to my test project, but it looks like my tests are not being executed by the CI build since they are not included in `SQLProvider.Tests.sln`.  
Outside of that, all my changes were isolated to `Providers.MsSqlServer.Ssdt.fs`, so it's not clear how that could cause a problem with the main tests.

Thorium:
The test runner picks assemblies by `"tests/**/bin/Release/*Tests*.dll"` and that seems to hit the MsSqlSsdt.Tests.
Should these tests be run on build or not?

Thorium:
Ok, it seems that the build doesn't find the unit tests anymore. 
This cannot be issue of this PR, it has to be some kind of update from internet, either to VisualStudio, Fake or Paket.

The test passed when I ran them from VisualStudio manually. 
So I released this to NuGet as SQLProvider version 1.1.100

I'm not too keen to spending time fixing the environment as I think we should move towards releasing from the net-standard branch, as SQLProvider 1.2.

JordanMarr:
I put the SSDT unit test project inside its own .sln so that I could reference the SSDT .sqlproj in it as well. I never added the new sln to the build, so I'm pretty sure the tests are not run during the CI build. 
I'll add it on the next round of changes.

PR #713, opened 2021-02-14
Separated dacpac parser to make it easier to resuse in other community projects

JordanMarr:
## Proposed Changes

Per our chat, I refactored the SSDT .dacpac parser into its own file, `Ssdt.DacpacParser.fs`. 
Any logic specific to SQLProvider has been removed from this file so it should be self-contained and easily reusable in other community projects (per your suggestion of using paket single file references).

I had previously created some FSharp.Core polyfills (`Array.last`, `Option.ofObj`, etc) which were moved into `Ssdt.Polyfills.fs`. This allows them to be usable by `Ssdt.DacpacParser.fs` without polluting the rest of the namespace.

## Types of changes

Refactor (moving & renaming only).


Thorium:
Great, thanks!

PR #715, opened 2021-03-08
Fix UseOptionTypes documentation

breyed:
## Proposed Changes

Documentation fix. Correct that a default value isn't always returned if UseOptionTypes is false. Also indicate default.

## Types of changes

## Checklist


Thorium:
Thanks!

PR #735, opened 2021-07-10
Added async operations to retrieve exactly one element from IQueryable.

b0wter:
## Proposed Changes

I am referring to a chat on [gitter](https://gitter.im/fsprojects/SQLProvider?at=60e8014a48f9b1084a1dd183).
I've added methods to `Seq` to retrieve exactly one entity from a query asynchronously. There are three new functions:

`Seq.exactlyOneAsync` - behaves like `Seq.exactleOne`, e.g. throws if the `seq` is empty or there is more than one element.
`Seq.exactlyOneOrDefaultAsync` - Returns the single element from the list and `Unchecked.defaultof<'T>` if there are no elements or more than one.
`Seq.exactlyOneOrNoneAsync` - Returns the single element as an option and `None` if the `seq` is emty or there is more than one element.

Here is an example:
```    
let findById (ctx: sql.dataContext) (entityId: Guid) : Async<sql.dataContext.``public.fooEntity`` option> =
    query {
        for foo in ctx.Public.Bar do
        where (foo.FooNo = entityId)
    } |> Seq.exactlyOneOrNoneAsync
```

## Types of changes

## Checklist

I was only able to build the library locally but not to run the tests. I have tested the changes in a side project of mine.

## Further comments

If you like this PR I'll add documentation as well. I think [this section](https://github.com/fsprojects/SQLProvider/blob/master/tests/SqlProvider.Tests/Readme.md#select-query-async) would be a good place to add a simple example and the method list [here](https://fsprojects.github.io/SQLProvider/core/async.html) would require an update..

### DIFF
--- a/src/SQLProvider.Runtime/SqlRuntime.Async.fs
+++ b/src/SQLProvider.Runtime/SqlRuntime.Async.fs
@@ -32,18 +32,21 @@ module AsyncOperations =
                         do! coll.EvaluateQuery()
                         return yieldseq en
                     | c -> return yieldseq en
-                else 
+                else
                     return yieldseq en
         }
 
-    let private fetchTakeOne (s:Linq.IQueryable<'T>) =
+    let private fetchTakeN (n: int) (s:Linq.IQueryable<'T>) =
         async {
             match s with
             | :? IWithSqlService as svc ->
-                return! executeQueryAsync svc.DataContext svc.Provider (Take(1, svc.SqlExpression)) svc.TupleIndex
+                return! executeQueryAsync svc.DataContext svc.Provider (Take(n, svc.SqlExpression)) svc.TupleIndex
             | c ->
                 return c :> Collections.IEnumerable
         }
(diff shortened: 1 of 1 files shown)


Thorium:
Thanks, looking generally good!
One thing though, F# version of `getExactlyOneOrNoneAsync` is called `tryExactlyOneAsync`
and I think there doesn't exist a method called `getExactlyOneOrDefaultAsync` because it's not clear if it should throw or not on too-many -case.

b0wter:
Ups, that's kind of embarrassing. I've renamed `(get)ExactlyOneOrDefaultAsync` to `(get)TryExactlyOne` and removed `getExactlyOneOrDefaultAsync` since `Unchecked.defaultof<'T>` does not seem very idiomatic to me and I think you're right in that it is too obtuse.

Thorium:
Thanks! Released in NuGet, version 1.2.6

PR #742, opened 2021-08-09
Azure Functions .dacpac Resolution Fix

varon:
## Proposed Changes

This simply searches a much wider set of locations to try and find a .dacpac. Doing some path magic (and it's best effort) to try and ensure that the path is resolved in wider circumstances.

Specifically, this solves the case where EntryAssembly points to a totally different location, such as when running under the Azure Functions harness.

## Types of changes

## Checklist

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

### DIFF
--- a/src/SQLProvider.Runtime/Providers.MsSqlServer.Ssdt.fs
+++ b/src/SQLProvider.Runtime/Providers.MsSqlServer.Ssdt.fs
@@ -5,6 +5,9 @@ open System
 open System.Collections.Concurrent
 open System.Data
 open System.Data.SqlClient
+open System.IO
+open System.Reflection
+open System.Text
 open FSharp.Data.Sql
 open FSharp.Data.Sql.Transactions
 open FSharp.Data.Sql.Schema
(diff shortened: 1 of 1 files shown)


Thorium:
Looks generally good to me. 
Just a question, should it somehow try-catch the ".." as there could be some possible issues, e.g.:

- What if there is no user permissions to go folders going parent of the root, ".." ?
- What if you have mapped the path to be a root (like Windows map network drive) and ".." doesn't exist?

varon:
Integrated all of your suggestions:
- Reduced path duplication
- Added a try-catch around the file opening (for permission issues)
- Handled badly formed paths by just replacing with '.'
- Also added a (likely the most useful) environment variable: `SQLPROVIDER_SSDT_DACPAC_FILE_LOOKUP_PATH`

Did some testing on network paths and local too. This should also work as expected.

Thorium:
Thanks, released in NuGet version 1.2.8

varon:
You absolute legend. Thank you!

PR #752, opened 2021-11-05
Rename command parameter to 'com'

flavluc:
## Proposed Changes

The `con` parameter name sounds misleading as it's usually given to an `IDbConnection` object. Since this is an `IDbCommand`, `com` is the right fit.


Thorium:
Thanks!

PR #766, opened 2022-06-23
ValueOption support

Thorium:
## Proposed Changes

This changes the parameter UseOptionTypes to be, instead of current true/false,
an DU:

- FSharp.Data.Sql.Common.NullableColumnType.NO_OPTION - database nulls are default (as current false)
- FSharp.Data.Sql.Common.NullableColumnType.OPTION - database nulls are `option<_>` (as current true)
- FSharp.Data.Sql.Common.NullableColumnType.VALUE_OPTION - database nulls are `voption<_>` (new feature)

Many of the database types are primitives and thus a struct would do fine.
This is not changing the SqlEntity class itself, so you can still update the database values and save datacontext as usual.

### DIFF
--- a/docs/content/core/async.fsx
+++ b/docs/content/core/async.fsx
@@ -17,7 +17,7 @@ type TypeProviderConnection =
         ConnectionString = "Server=localhost;Database=test;User=test;Password=test",
         DatabaseVendor = Common.DatabaseProviderTypes.MSSQLSERVER,
         IndividualsAmount=1000,
-        UseOptionTypes=FSharp.Data.Sql.Common.OPTION, 
+        UseOptionTypes=FSharp.Data.Sql.Common.NullableColumnType.OPTION, 
         CaseSensitivityChange = Common.CaseSensitivityChange.ORIGINAL>
 
 (**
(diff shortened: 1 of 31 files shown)


Thorium:
By my initial testing, the `ValueOption<_>` used 50% less CPU and 10% less memory than `Option<_>`
which might be relevant if you have a lot of things going on.

The only restriciton is that ValueOption cannot be compared nicely (this is due to FSharp, not SQLProvider):

```fsharp
query {
       for emp in ctx.Dbo.Employees do
       where (emp.FirstName = ValueSome("Steven"))
       select emp.FirstName
} // System.InvalidOperationException: The binary operator Equal is not defined for the types
```

But you can do of course e.g.:

```fsharp
query {
       for emp in ctx.Dbo.Employees do
       where (emp.FirstName.IsSome && emp.FirstName.Value = "Steven")
       select emp.FirstName
} // works
```

...and if you absolutely want to compare 2 voptions, you can wrap them e.g. like this:

```fsharp
query {
       for emp in ctx.Dbo.Employees do
       where (Some(emp.FirstName) = Some(ValueSome("Steven")))
       select emp.FirstName
}
```

Issue #776, opened 2022-11-17
Queries failing for schemas / tables that contain special characters - MySQL


Choc13:
**Describe the bug**
We have a schema name like `my-schema` in MySQL. When querying this schema we are required to enclose it in backticks to escape the hyphen. It seems however that the SQL generated by SQL Provider does not contain backticks around schema or table names (although it does around column names). 

In principle would it be possible to always wrap schema and table names in backticks to ensure they are properly escaped? If so I'm happy to submit a PR. 

On a side note I think this might be the underlying cause of the other issue (#773) I opened about the missing `Create(...)`  functions.

**To Reproduce**
Create a schema in MySQL that contains a `-` in the name and try to query it. 

**Expected behavior**
The generated SQL should be more like:

```sql
SELECT `a` as `a`
FROM `my-schema`.`my-table`
```

Thorium:
Closed as fixed.

PR #777, opened 2022-12-03
Always quote schema name with backticks in MySQL queries.

Choc13:
## Proposed Changes

Adds backtick quotes around all schema names in MySQL queries. Addresses #776.

## Types of changes

## Checklist

- [x] Build and tests pass locally (Build passes, but I wasn't able to get all the tests to run locally yet. Will keep trying.)

### DIFF
--- a/src/SQLProvider.Runtime/Providers.MySql.fs
+++ b/src/SQLProvider.Runtime/Providers.MySql.fs
@@ -23,9 +23,9 @@ module MySql =
 
     let findType name =
         match assembly.Value with
-        | Choice1Of2(assembly) -> 
-            let types = 
-                try assembly.GetTypes() 
+        | Choice1Of2(assembly) ->
+            let types =
+                try assembly.GetTypes()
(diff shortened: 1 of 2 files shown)


Thorium:
Thanks, this is definitely improvement.

Do you think the replace is too naive? (I know it was there already.) Like what if your table/column contains a weird name like `Person[backup]` ?

Choc13:
That is a good point. I wondered the same but didn't change it as I saw a comment about `[]` being used as an internal representation so figured the replacement was there in order to switch to the MySQL specific quotations. However I think you are right that if the table/schéma did also contain those chars then they would be incorrectly substituted. 

It seems like a separate change set though to remodel the internal quoting behaviour to address this. I would have thought it should be sufficient to not use any quote symbol internally and just do vendor specific quoting in each provider, but there are likely reasons I'm not aware of given I've only spent an hour or so looking over this code.

Thorium:
You are correct, thanks, will release this.

PR #779, opened 2023-01-07
Add user defined data type support for SQL Server SSDT

MMagueta:
## Proposed Changes

Currently, SSDTs for SQL Server do not support user defined data types. This means if a simple custom type is defined, such as the one from below, the column does not get translated to F#.

A simple UDDT:

```tsql
CREATE TYPE LATITUDE FROM REAL;
GO
CREATE
    RULE RuleLatitudeRange
    AS
    @per >= -180.0 AND @per <= 180.0
GO
EXEC SP_bindrule 'RuleLatitudeRange', 'LATITUDE'
GO
```

From the lsp:
Screenshot 2023-01-07 at 01 03 14

On the following table, where all the types are also simple overloads on REAL and INT:
```tsql
CREATE TABLE dbo.TestUDDTs
(
    ID INT PRIMARY KEY,
    A MINUTES NOT NULL,
    B COMPASS NOT NULL,
    C LATITUDE NOT NULL,
    D LONGITUDE NOT NULL,
    E DAYS NOT NULL,
);
```

With this PR we now have:
Screenshot 2023-01-07 at 01 12 33

A further explanation on the implementation may be found under `Further comments`

## Types of changes

Changes this PR introduces:

- [X] New feature (non-breaking change which adds functionality)

## Checklist

## Further comments

- The solution is:
  - Parse the nodes in XML:
```xml
<Element Type="SqlUserDefinedDataType" Name="[dbo].[MINUTES]">
	<Relationship Name="Schema">
		<Entry>
			<References ExternalSource="BuiltIns" Name="[dbo]" />
		</Entry>
	</Relationship>
	<Relationship Name="Type">
		<Entry>
			<References ExternalSource="BuiltIns" Name="[int]" />
		</Entry>
	</Relationship>
</Element>
```
  - Generate a Map<UDDTName, UDDTInheritedType> (both algebraic aliases for string), where `UDDTName` is the given name to the type, and  `UDDTInheritedType` is the base type that `UDDTName` inherits from
  - Add a new parameter to `tryFindMappingOrVariant` where such map is passed
  - Attempts to find the type in the normal `typeMappingsByName` map, if it does not find, attempts once again on the UDDT map, if something was found, searches for the base type in `typeMappingsByName`, otherwise defaults to `SQL_VARIANT` and consequently to `System.Object`
- Since UDDTs can only be referenced to simple primitives in SQL Server, I leverage the fact I don't have to worry on UDDTs being defined in terms of another UDDT.

### DIFF
--- a/src/SQLProvider.Runtime/Providers.MsSqlServer.Ssdt.fs
+++ b/src/SQLProvider.Runtime/Providers.MsSqlServer.Ssdt.fs
@@ -170,8 +170,8 @@ module MSSqlServerSsdt =
         match typeMappingsByName.TryFind dataType with
         | Some tm -> tm
         | None ->
-            match Map.tryFind dataType uddts with
-            | Some x -> tryFindMappingOrVariant uddts x
+            match Map.tryFind (UDDTName dataType) uddts with
+            | Some (UDDTInheritedType x) -> tryFindMappingOrVariant uddts x
             | None -> (typeMappingsByName.TryFind "SQL_VARIANT").Value
 
     let ssdtTableToTable (tbl: SsdtTable) =
--- a/src/SQLProvider.Runtime/Providers.MsSqlServer.Ssdt.fs
+++ b/src/SQLProvider.Runtime/Providers.MsSqlServer.Ssdt.fs
@@ -165,32 +165,31 @@ module MSSqlServerSsdt =
     let tryFindMapping (dataType: string) =
         typeMappingsByName.TryFind (dataType.ToUpper())
 
-    let rec tryFindMappingOrVariant (uddts: SsdtUserDefinedDataType list) (dataType: string) =
-        match typeMappingsByName.TryFind (dataType.ToUpper()) with
+    let rec tryFindMappingOrVariant (uddts: SsdtUserDefinedDataType) (dataType: string) =
+        let dataType = dataType.ToUpper()
+        match typeMappingsByName.TryFind dataType with
         | Some tm -> tm
         | None ->
-            match List.tryFind (fun x -> x.Name = dataType) uddts with
-            | Some x -> tryFindMappingOrVariant uddts x.Inheritance
+            match Map.tryFind dataType uddts with
+            | Some x -> tryFindMappingOrVariant uddts x
             | None -> (typeMappingsByName.TryFind "SQL_VARIANT").Value
 
     let ssdtTableToTable (tbl: SsdtTable) =
         { Schema = tbl.Schema ; Name = tbl.Name ; Type =  if tbl.IsView then "view" else "base table" }
 
-
     let ssdtColumnToColumn uddts (tbl: SsdtTable) (col: SsdtColumn) =
-        let typeMapping = tryFindMappingOrVariant uddts col.DataType
-        Some
-            { Column.Name = col.Name
-              Column.TypeMapping = typeMapping
-              Column.IsNullable = col.AllowNulls
-              Column.IsPrimaryKey =
-                tbl.PrimaryKey
-                |> ValueOption.map (fun pk -> pk.Columns |> List.exists (fun pkCol -> pkCol.Name = col.Name))
-                |> ValueOption.defaultValue false
-              Column.IsAutonumber = col.IsIdentity
-              Column.HasDefault = col.HasDefault
-              Column.IsComputed = col.ComputedColumn
-              Column.TypeInfo = if col.DataType = "" then ValueNone else ValueSome col.DataType }
+        let typeMapping = tryFindMappingOrVariant uddts col.DataType        
+        { Column.Name = col.Name
+          Column.TypeMapping = typeMapping
+          Column.IsNullable = col.AllowNulls
+          Column.IsPrimaryKey =
+            tbl.PrimaryKey
+            |> ValueOption.map (fun pk -> pk.Columns |> List.exists (fun pkCol -> pkCol.Name = col.Name))
+            |> ValueOption.defaultValue false
+          Column.IsAutonumber = col.IsIdentity
+          Column.HasDefault = col.HasDefault
+          Column.IsComputed = col.ComputedColumn
+          Column.TypeInfo = if col.DataType = "" then ValueNone else ValueSome col.DataType }
 
 type internal MSSqlServerProviderSsdt(tableNames: string, ssdtPath: string) =
     let schemaCache = SchemaCache.Empty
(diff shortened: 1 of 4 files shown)


src/SQLProvider.Runtime/Providers.MsSqlServer.Ssdt.fs
```diff
@@ -165,32 +165,31 @@ module MSSqlServerSsdt =
...
-        | None -> (typeMappingsByName.TryFind "SQL_VARIANT").Value
+        | None ->
+            match Map.tryFind (UDDTName dataType) uddts with
+            | Some (UDDTInheritedType x) -> tryFindMappingOrVariant uddts x
+            | None -> (typeMappingsByName.TryFind "SQL_VARIANT").Value
 
     let ssdtTableToTable (tbl: SsdtTable) =
         { Schema = tbl.Schema ; Name = tbl.Name ; Type =  if tbl.IsView then "view" else "base table" }
 
-    let ssdtColumnToColumn (tbl: SsdtTable) (col: SsdtColumn) =
```
MMagueta, line 176: Hey @Thorium, about this function, I was doubtful if I should remove the Option from it since I am unaware of the consequences it might lead into in the `GetColumns` member with the List.choose. In any case, I removed it in favor of the UDDT map

JordanMarr:
Nice. 👍

MMagueta:
I have added a simple test, but I am having problems with running them locally on my machine. If someone can build them I would appreciate.

I tested using this script with a database I defined myself:
```fsharp
open FSharp.Data.Sql
open FSharp.Data.Sql.Common

let [<Literal>] vendor = DatabaseProviderTypes.MSSQLSERVER_SSDT
type Schema = SqlDataProvider<vendor, SsdtPath="/Users/mmagueta/Projects/FOSS/TestSQLProvider/base.dacpac">

type Config =
    { Host: string
      Port: string
      Database: string
      User: string
      Password: string }

[<EntryPoint>]
let main _ = 
    let config = {Host = "localhost"; Port = "1433"; Database = "msdb"; User = "sa"; Password = "P@ssw0rd"}
    let dataCtx =
        Schema.GetDataContext($"TrustServerCertificate = true; Server={config.Host},{config.Port}; Initial Catalog={config.Database}; User Id={config.User}; Password={config.Password}")
    let entity = dataCtx.Dbo.TestUddTs.Create()
    printfn "%A" <| ((entity.A.GetType().FullName) = "System.Int32")
    0
```

src/SQLProvider.Runtime/Providers.MsSqlServer.Ssdt.fs
Thorium: I'm not familiar with this, @JordanMarr knows better. Maybe the idea was that totally unknown column types are just skipped by SQLProvider. I would expect that e.g. stored procedures could return something unknown that is then just treated as an obj.

JordanMarr: It was an arbitrary decision on my part to omit the column instead of adding as a variant / `obj`.
Your change to call `tryFindMappingOrVariant` seems ok to me.

MMagueta:
Anything I would still have to go through? If it's ok, I am done with the scope of this PR @JordanMarr

src/SQLProvider.Runtime/Providers.MsSqlServer.Ssdt.fs
```diff
@@ -165,32 +165,31 @@ module MSSqlServerSsdt =
     let tryFindMapping (dataType: string) =
         typeMappingsByName.TryFind (dataType.ToUpper())
 
-    let tryFindMappingOrVariant (dataType: string) =
-        match typeMappingsByName.TryFind (dataType.ToUpper()) with
+    let rec tryFindMappingOrVariant (uddts: SsdtUserDefinedDataType) (dataType: string) =
+        let dataType = dataType.ToUpper()
+        match typeMappingsByName.TryFind dataType with
         | Some tm -> tm
-        | None -> (typeMappingsByName.TryFind "SQL_VARIANT").Value
```
JordanMarr, line 171: While you're in there, can you improve this line so that it doesn't call `TryFind` and `Value` (I was probably rushing):

```F#
| None -> typeMappingsByName["SQL_VARIANT"]
```

MMagueta: Sure thing, the map should always have it at this point, so makes sense.

Changed it on this commit:
https://github.com/fsprojects/SQLProvider/pull/779/commits/f0592016fed253502e93ec6802b2e431a17dd47e

JordanMarr:
I have submitted a PR to add my old SSDT unit test project back to the test sln.
If @Thorium wants to add this back, I think it could be very useful way to add some test coverage around the SSDT provider, including your changes.

src/SQLProvider.Runtime/Ssdt.DacpacParser.fs
```diff
@@ -386,6 +390,28 @@ let parseXml(xml: string) =
             Description = description
         }
 
+    let parseUserDefinedDataType (udts: XmlNode) =
```
ratsclub, line 393:
```suggestion
    let parseUserDefinedDataType (uddts: XmlNode) =
```

src/SQLProvider.Runtime/Ssdt.DacpacParser.fs
```diff
@@ -386,6 +390,28 @@ let parseXml(xml: string) =
...
+        let name = udts |> att "Name"
```
ratsclub, line 394:
```suggestion
        let name = uddts |> att "Name"
```

src/SQLProvider.Runtime/Ssdt.DacpacParser.fs
```diff
@@ -386,6 +390,28 @@ let parseXml(xml: string) =
...
+        let name = name |> RegexParsers.splitFullName |> fun parsed -> String.Join(".", parsed) |> fun n -> n.ToUpper() |> UDDTName
+        let inheritedType =
+            udts
```
ratsclub, line 397:
```suggestion
            uddts
```

Thorium:
Thanks!

Thorium:
I'll release new version tomorrow.

Thorium:
> Since UDDTs can only be referenced to simple primitives in SQL Server

You can do user-defined table type like

```sql
CREATE TYPE [F#].[Payment] AS TABLE(
	[Date] [smalldatetime] NULL,
	[Amount] [money] NULL
)
```

MMagueta:
> > Since UDDTs can only be referenced to simple primitives in SQL Server
> 
> You can do user-defined table type like
> 
> ```sql
> CREATE TYPE [F#].[Payment] AS TABLE(
> 	[Date] [smalldatetime] NULL,
> 	[Amount] [money] NULL
> )
> ```
> 
> Although I'm not sure if SQL is a good tool for this kind of domain modelling as it's not meant to be a general purpose programming language.

This is somewhat useful for hidden behavior inside procedures (to group a query and insert somewhere from it), triggers (like error handling to persist a rollback), etc; but not so much to expose it via the provider. That's why I defined a user defined data type only, UDTTs are not even being parsed on the XML from the dacpac.

PR #790, opened 2023-04-26
Add OnConflict handling for MySQL.

Choc13:
## Proposed Changes

I'd like to propose adding support for `OnConflict` in the `MySQL` provider. I believe it can be handled via the `ON DUPLICATE KEY UPDATE` clause that is available in MySQL. 

## Types of changes

## Checklist

## Further comments

I've not written any tests yet as I wanted to check that you open to accepting this new feature first and to also ask a few questions about the implementation, which I will point out in comments on the changes.

### DIFF
--- a/src/SQLProvider.Runtime/Providers.MySql.fs
+++ b/src/SQLProvider.Runtime/Providers.MySql.fs
@@ -376,7 +376,7 @@ type internal MySqlProvider(resolutionPath, contextSchemaPath, owner:string, ref
         let cmd = (this :> ISqlProvider).CreateCommand(con,"")
         cmd.Connection <- con
 
-        let columnNames, values =
+        let columnNamesWithValues =
             (([],0),entity.ColumnValues)
             ||> Seq.fold(fun (out,i) (k,v) ->
                 let name = sprintf "@param%i" i
@@ -385,13 +385,25 @@ type internal MySqlProvider(resolutionPath, contextSchemaPath, owner:string, ref
             |> fun (x,_)-> x
             |> List.rev
             |> List.toArray
-            |> Array.unzip
+
+        let columnNames, values = columnNamesWithValues |> Array.unzip
 
         sb.Clear() |> ignore
-        ~~(sprintf "INSERT INTO %s (%s) VALUES (%s); SELECT LAST_INSERT_ID();"
+        ~~(sprintf "INSERT INTO %s (%s) VALUES (%s)"
             (entity.Table |> quotedTableName)
             ("`" + (String.Join("`, `",columnNames)) + "`")
             (String.Join(",",values |> Array.map(fun p -> p.ParameterName))))
+
+        match entity.OnConflict with
+        | Throw -> ()
+        | Update ->
+          ~~(sprintf " ON DUPLICATE KEY UPDATE %s"
+                (String.Join(",", columnNamesWithValues |> List.map(fun (c,p) -> sprintf "`%s`=%s" c p.ParameterName))))
+        | DoNothing ->
+          ~~(sprintf " AS x ON DUPLICATE KEY UPDATE %s"
+                (String.Join(",", columnNamesWithValues |> List.map(fun (c,p) -> sprintf "`%s`=x.`%s`" c p.ParameterName))))
+
+        ~~"; SELECT LAST_INSERT_ID();"
 
         values |> Array.iter (cmd.Parameters.Add >> ignore)
         cmd.CommandText <- sb.ToString()
--- a/src/SQLProvider.Runtime/Providers.MySql.fs
+++ b/src/SQLProvider.Runtime/Providers.MySql.fs
@@ -398,10 +398,10 @@ type internal MySqlProvider(resolutionPath, contextSchemaPath, owner:string, ref
         | Throw -> ()
         | Update ->
           ~~(sprintf " ON DUPLICATE KEY UPDATE %s"
-                (String.Join(",", columnNamesWithValues |> List.map(fun (c,p) -> sprintf "`%s`=%s" c p.ParameterName))))
+                (String.Join(",", columnNamesWithValues |> Array.map(fun (c,p) -> sprintf "`%s`=%s" c p.ParameterName))))
         | DoNothing ->
           ~~(sprintf " AS x ON DUPLICATE KEY UPDATE %s"
-                (String.Join(",", columnNamesWithValues |> List.map(fun (c,p) -> sprintf "`%s`=x.`%s`" c p.ParameterName))))
+                (String.Join(",", columnNamesWithValues |> Array.map(fun (c,_) -> sprintf "`%s`=x.`%s`" c c))))
 
         ~~"; SELECT LAST_INSERT_ID();"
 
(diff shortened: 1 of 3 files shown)


src/SQLProvider.Runtime/Providers.MySql.fs
```diff
@@ -385,14 +385,26 @@ type internal MySqlProvider(resolutionPath, contextSchemaPath, owner:string, ref
...
             ("`" + (String.Join("`, `",columnNames)) + "`")
             (String.Join(",",values |> Array.map(fun p -> p.ParameterName))))
 
+        match entity.OnConflict with
+        | Throw -> ()
+        | Update ->
+          ~~(sprintf " ON DUPLICATE KEY UPDATE %s"
+                (String.Join(",", columnNamesWithValues |> List.map(fun (c,p) -> sprintf "`%s`=%s" c p.ParameterName))))
+        | DoNothing ->
+          ~~(sprintf " AS x ON DUPLICATE KEY UPDATE %s"
```
Choc13, line 403: I think this `AS x` syntax might only be supported since MySQL 8. We could instead use the `VALUES()` syntax, but that is now deprecated. I have written this such that it would only be an issue if a user was to specify `OnConflict.DoNothing` as the `AS x` part is conditional on that. So this shouldn't regress any existing code given that `OnConflict` is currently not supported for `MySQL` anyway. 

If there is a way to determine the MySQL version then I could use that to generate different queries based on the version, but for now this seemed like the most pragmatic approach.

src/SQLProvider.Runtime/Providers.MySql.fs
```diff
@@ -385,14 +385,26 @@ type internal MySqlProvider(resolutionPath, contextSchemaPath, owner:string, ref
...
+                (String.Join(",", columnNamesWithValues |> List.map(fun (c,p) -> sprintf "`%s`=x.`%s`" c p.ParameterName))))
+
+        ~~"; SELECT LAST_INSERT_ID();"
```
Choc13, line 406: From my testing on a MySQL instance this will return `0` if a row is updated instead of inserted. I tried to follow the code through to see where this value is used and my understanding is that an attempt is made to set the ID of the entity that was just created. But it looks like there is some code there to check whether the ID is already set, in which case this should be fine as we're only returning `0` in the case where it was an update and at that point it would have already existed anyway.

But please let me know if my understanding of that part of the code is wrong.

Thorium:
OnConflict is definitely not my favourite thing (I prefer transactions) but yes PRs are accepted!

src/SQLProvider.Runtime/Providers.MySql.fs
Thorium: The negative side of detecting version is that it's an extra network call to the db.

Thorium:
Is this work in progress, or waiting for merge?

Choc13:
Hi, sorry I was busy this last week. It sounds like you're happy to accept the MR, so let me see if I can add some tests if possible. 

The build seems to have failed for a transient reason that I didn't understand. Could you advise on how to fix that? Thanks 👍

src/SQLProvider.Runtime/Providers.MySql.fs
Choc13: Yes, I think I'll leave it as this then and it will just not be supported on older versions of MySQL as it currently isn't.

Choc13:
> OnConflict is definitely not my favourite thing (I prefer transactions) but yes PRs are accepted!

Curious what you meant by "I prefer transactions" here? How would you typically handle upserting rows? Fetching everything that matches and then doing an update on that entity if it already exists or creating a new one if it doesn't?

Thorium:
Off-topic but as I do financial systems, I try to avoid updates and upserts as it might lose some important historical data. I'd rather just insert a new row and have some kind of order/timestamp to indicate what is the latest and currently valid row. That way I can ask how the system was looking back in time.

Choc13:
I'm unable to get the build or tests running locally and the AppVeyor build also seems to be failing to setup the SQLServer instance. I'm not too sure how to fix that issue. I can try and investigate, but if you were able to point me in the right direction that would be great. Thanks.

Thorium:
I'm not very familiar with the test-projects, but definitely as this is MySQL only the MySQL related errors should be relevant and a "System.Data.SqlClient.SqlException (0x80131904): A network-related or instance-specific " is not relevant for this PR.

Choc13:
> Off-topic but as I do financial systems, I try to avoid updates and upserts as it might lose some important historical data. I'd rather just insert a new row and have some kind of order/timestamp to indicate what is the latest and currently valid row. That way I can ask how the system was looking back in time.

Ah yeah, I do that in most places, but in a few places an upsert is all that is needed.

Choc13:
> 

OK, do you recommend that I comment out the SQLServer tests for the time being then, make sure it passes without them and then uncomment them before merging?

Thorium:
> > 
> 
> OK, do you recommend that I comment out the SQLServer tests for the time being then, make sure it passes without them and then uncomment them before merging?

That's fine too, but why you cannot get the build working locally?
I'd be keen to update the docs or something to help people to get local build working.

Choc13:
Ah, it was because I didn't have mono installed. It's getting further now.

Choc13:
I know you said you don't know much about the test project, but do you know if the `scipts/*.fsx` test files are supposed to run when invoking `dotnet fake run build.fsx -t RunTests`. I can't seem to get them to trigger. I see some references to setting up a Postgres instance in `build.fsx`, but there is no mention of setting up a MySQL instance in there. There is a docker file and some scripts that look like they would initialise a MySQL instance, but I can't seem them being invoked anywhere in the build. 

Is the intention that those test scripts are setup and run manually by starting the docker container and running the `fsx` file?

Thorium:
The basic test-fsproj executes only .fs files.

The .fsx files are rather for testing manually. They typically need some drivers or connection string settings tweaked first.

I know people have set-up dockers to ease testing with different databases. I'm not so familiar with those. I'd expect there are some .fs files with confitional compilation to execute that part on CI only. But I don't know which databases are included. 

SQLite tests run without docker.

src/SQLProvider.Runtime/Providers.MySql.fs
Choc13: My tests revealed that this was actually doing the wrong thing anyway, the `AS x` syntax is a way to refer to the values being inserted, not the existing column values as I had intended. So the correct implementation is simply to do `ON DUPLICATE KEY UPDATE col1 = col1, col2 = col2` etc which I believe is supported on older versions of MySQL too. 👍

Choc13:
I'm happy for this to be merged now if you are.

Thorium:
Thanks!

Thorium:
Umm, what is this DoNothing: 
```columnNamesWithValues |> Array.map(fun (c,_) -> sprintf "`%s`=`%s`" c c```

Also, if you have multiple columns, then are you missing spaces in the string?

Choc13:
> 

It is what I mentioned above. It generates something like the following `ON DUPLICATE KEY UPDATE col1 = col1, col2 = col2`. Which means just set the column to its existing value already in the table. The `String.join` takes care of comma separating each `col=col` so I don't think spaces are necessary. I tested this case locally and also inspected the SQL it generated and it looked fine, even with multiple columns in the query.

PR #797, opened 2023-07-09
SubmitUpdates ordering

### LINKED ISSUE #758: SubmitUpdates Operation Order
**Is your feature request related to a problem? Please describe.**
When inserting a record with child records, there is no guarantee that the child records will be created after the parent. This causes a database error.

**Describe the solution you'd like**
It would be good if there was some consistency in the order of the operations. Something like FIFO. Perhaps the data SubmitUpdates method could accept a ProcessingStrategy input.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

gbtb:
Example of one possible way to fix #758

### DIFF
--- a/docs/content/core/constraints-relationships.fsx
+++ b/docs/content/core/constraints-relationships.fsx
@@ -10,4 +10,7 @@ constraint in the database.
 
 You can gain access to these child or parent entities by simply enumerating 
 the property in question.
+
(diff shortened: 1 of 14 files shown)


Thorium:
I think we want sort to SqlRuntime.DataContext.fs here too:
```fsharp
member __.GetPendingEntities() = pendingChanges.Keys |> Seq.toList
```

That is only used for possibly manually asking changes: dbContext.GetUpdates(). Most cases would be e.g. logging or debugging: should you do SubmitUpdates() or ClearUpdates() next.

gbtb:
I think it's ready for review. I would throw in couple of tests for a good measure, but I can't build tests projects no matter what I try. It struggles to find SQLite interop dlls, probably because of some weird linux+mono+472 interactionj.

Thorium (approved):
Looks good to me

PR #827, opened 2024-08-06
Remove globals.json and build with NET 8.0 compiler

Thorium:
Initial testing...


Thorium:
This is definitely not problematic to compile, even with latest FAKE 6.1.1 updates. 
But I'll push this to NuGet as 1.4.0-alpha for some further testing:
https://www.nuget.org/packages/SQLProvider/1.4.0-alpha

PR #846, opened 2025-06-03
Optimizations for MapTo function

QrystaL:
Seems like the initial issue happened only for `ValueOption` (struct). We may use `RuntimeHelpers.GetUninitializedObject` in that case, however it's not available in .net standard 2.0


QrystaL:
Minimum repro:

`let a1: obj option = None
let b1 = box a1 :?> string option // OK

let a2: obj voption = ValueNone
let b2 = box a2 :?> string voption // InvalidCastException`

Thorium:
Makes sense to restrict only to FSharpValueOption. But we do need to still support .NET Standard 2.0.

I think the major problem with MapTo is that it actually does select * and then mapto, meanwhile, as the entity is pre-known, what it could do (if within query { ... }), is to first read the required columns and then select only those. Of course then there might be edge cases like "select (cust.Id, cust.MapTo...)"

